PR #3, opened 2021-11-16
Initial skeleton dock compose

randradas:
**Description of the change**

This changes introduces a first version of a docker-compose.yaml that can be used to deploy Carto dependencies ecosystem. This version uses fake nodejs containers to simulate Carto software. The goal is to have the Carto dependencies (nginx, postgresql, etc) ready to be deployed and an easy way to substitute fake container with Carto software.

**Benefits**

Get Carto dependencies ready to be deployed even though the Carto software is not included into the nodejs containers.


randradas:
I still need to ask for some containers that could be needed before merging, so expect changes in the number of nodejs containers.

PR #4, opened 2021-11-16
Run linters as status check on PRs

randradas:
**Description of the change**

Set linters to run during PR as status checks

**Benefits**

PR are verified by GitHub workflows


randradas:
Expect more commits with some linters that should be run during each PR

PR #5, opened 2021-11-17
Adding linters to the PR-verify workflow

randradas:
**Description of the change**

This change will set current workflow as part of the verification process required to merge a PR. Also, the workflow need to include a minimal set of checks so linters have been added.

**Benefits**

PRs are verified by linters


.github/workflows/lint-codebase.yaml
```diff
@@ -0,0 +1,20 @@
...
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+      - run: echo "🐧 This job is now running on a ${{ runner.os }} server hosted by GitHub!"
+      - run: echo "🔎 The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          # Full git history is needed to get a proper list of changed files within `super-linter`
+          fetch-depth: 0
+      - name: Lint Code Base
+        uses: github/super-linter@v4
```
javsalgar, line 16: I checked the super linter but it doesn't have any helm linter specifics apart from kubeval. I think that we should add a step that at least installs helm so we can run `helm lint`, and `helm template` so we can render yaml and apply the `yamllint` linter, as well as kubeval (we have some hooks in the bitnami/charts repo that perform this action). In addition to this, it would be valuable to have the `readme-generator-for-helm` so we can also check that all the metadata that we are defining in the charts is correct (maybe that's outside the scope of linting and maybe another step/workflow should be added). I don't have much experience with how these workflows are configured, but with this feedback we could perform a more thorough lint.

javsalgar (commented):
The initial setup looks good to me, but make sure to check my comment

.github/workflows/lint-codebase.yaml
randradas: I added [Chart Testing](https://github.com/helm/chart-testing) also known as `ct` which run 'helm lint', version checking, YAML schema validation on 'Chart.yaml', YAML linting on 'Chart.yaml' and 'values.yaml', and maintainer validation.

However `ct` is not able to run render the template without installing it on a kind cluster so I need to find a way to run "helm template" only if the chart has been modify to avoid unnecessary steps.

randradas: Well. This what I've done is:
* Add _Super-Lint_ to lint the whole repo except the `chart/` directory which includes non-pure `.yaml` files like Helm chart templates that cause errors during YAML linting.
* Add _Chart Testing_ which is able to perform basic linting over Helm chart templates on `chart/` directory. This tool includes capabilities like deploying the chart on a _kind_ cluster but it are not used so it is not the scope right now.

Other requirements will be added in the future according with requirements that will appear.

randradas: Finally _Chart Testing_ tool is removed to keep it simple.

javsalgar (approved):

PR #7, opened 2021-11-25
Feature/initial skeleton dock compose

randradas:
**Description of the change**

This is a very initial docker-compose skeleton used to figure out how the Carto docker-compose.yaml would be. It helps to know how the nodejs container works which will be used to run Carto app.

This docker-compose uses fake images to run a Hello world application with node and also serve an static imagen which are bot things that need to be done to run Carto. Finally it is worth to mention that either the docker-compose file and the nginx.conf file are intentionally similar to the one provided by Carto.

**Benefits**

Helps to understand the architecture and can be used as sandbox to try some possible configurations.


docker/docker-compose.yaml
```diff
@@ -0,0 +1,100 @@
...
+    environment:
+      - BITNAMI_DEBUG=true
+
+  workspace-api-fake:
+    tty: true # Enables debugging capabilities when attached to this container.
+    image: docker.io/bitnami/node:16
+    depends_on:
+    - workspace-postgresql
+    - redis
+    command: sh -c 'npm install && npm start'
```
javsalgar, line 26: As the volume is shared between multiple applications, there could be issues with running multiple npm installs in parallel. Wouldn't it make more sense to change it to a `build: .` approach and create an image with the fake application?

docker/docker-compose.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - ./nodejs-fake-app/:/app
+
+  workspace-postgresql:
+    image: docker.io/bitnami/postgresql:13
+    volumes:
+      - 'postgresql_data:/bitnami/postgresql'
+    environment:
+      - POSTGRESQL_DATABASE=carto3_workspace
+      - POSTGRESQL_USERNAME=carto3
+      - POSTGRESQL_PASSWORD=
```
javsalgar, line 84: This line should not be necessary

javsalgar (changes requested):

docker/docker-compose.yaml
randradas: You are right, please let me explain the contest. Since this is just fake and the goal was to write a Carto-ish docker-compose file able to get node based container up I have push a commit that uses separate volumes, it is not clean enough but it is just a proof of concept docker-compose.yaml that I wrote while the actual Carto containers were not available to me due to legal paperwork in progress.

In the following commits in other PRs there will a better approach docker-compose that uses images that contains the application built-in and where volumes are not used (not persistence needed). Thank you so much for your feedback, it is useful, I didn't realize the volumes were being shared ;).

javsalgar (approved):

PR #8, opened 2021-12-14
Carto Helm chart, first cut.

randradas:
**Description of the change**

Carto 3 Helm Chart
**Benefits**

Be able to deploy Carto3 onprem solution on kubernetes


chart/Chart.yaml
```diff
@@ -0,0 +1,28 @@
...
+home: https://github.com/cartodb/carto3-helm/tree/stable/chart
+keywords:
+  - carto
+  - data
+  - spatial
+  - analysis
+name: carto
+sources:
+  - https://carto.com/
+version: 1.0.0
```
juan131, line 28: Shouldn't this be `0.0.1`?

chart/README.md
```diff
@@ -0,0 +1,1910 @@
+# carto
+
+[The CARTO Helm chart](https://github.com/CartoDB/carto-selfhosted) deploys CARTO in a self hosted environment.
+
+## TL;DR
+
+```console
+# TODO: Chart repository to be replaced by the Carto team
```
juan131: Is this still a TODO?

chart/README.md
```diff
@@ -0,0 +1,1910 @@
...
+
+## Introduction
+
+This chart bootstraps a [CARTO self hosted](https://github.com/CartoDB/carto-selfhosted) Deployment in a [Kubernetes](https://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- Helm 3.1.0
+- PV provisioner support in the underlying infrastructure
```
juan131: This isn't aligned with the README.md in the parent folder that says:
```
### Prerequisites

- Kubernetes 1.21.2
- Docker 20.10.8
- Helm 3.6.3
```

chart/README.md
```diff
@@ -0,0 +1,1910 @@
...
+| `workspaceSubscriber.livenessProbe.enabled`                 | Enable livenessProbe on workspace-subscriber containers                                                       | `true`                                 |
+| `workspaceSubscriber.livenessProbe.initialDelaySeconds`     | Initial delay seconds for livenessProbe                                                                       | `10`                                   |
+| `workspaceSubscriber.livenessProbe.periodSeconds`           | Period seconds for livenessProbe                                                                              | `30`                                   |
+| `workspaceSubscriber.livenessProbe.timeoutSeconds`          | Timeout seconds for livenessProbe                                                                             | `5`                                    |
+| `workspaceSubscriber.livenessProbe.failureThreshold`        | Failure threshold for livenessProbe                                                                           | `5`
```
juan131: Remove this:
```suggestion
```

juan131: Remove this line

juan131: I don't this parameter exists.. It'd be sth like `XXX.livenessProbe.enabled` instead

juan131:
```suggestion
In case you want to add extra environment variables (useful for advanced operations like custom init scripts), you can use the `*.extraEnvVars` property. For instance, to add extra environment variables on lds-api containers:

Comment (juan131) [inline chart/README.md]:
```suggestion
Alternatively, you can use a ConfigMap or a Secret with the environment variables. To do so, use the `*.extraEnvVarsCM` or the `*.extraEnvVarsSecret` values.
```

juan131:
```suggestion
If additional containers are needed in the any component's pods (such as additional metrics or logging exporters), they can be defined using the `*.sidecars` parameter. If these sidecars export extra ports, extra port definitions can be added using the `*.service.extraPorts` parameter.
```

juan131:
```suggestion
As an alternative, use one of the preset configurations for pod affinity, pod anti-affinity, and node affinity available at the [bitnami/common](https://github.com/bitnami/charts/tree/master/bitnami/common#affinities) chart. To do so, set the `*.podAffinityPreset`, `*.podAntiAffinityPreset`, or `*.nodeAffinityPreset` parameters.
```

juan131:
```suggestion
There are cases where you may want to deploy extra objects, such a ConfigMap containing your app's configuration or some extra deployment with a micro service used by your app. For covering this case, the chart allows adding the full specification of other objects using the `extraDeploy` parameter.
```

juan131 (changes requested):
Docs & metadata review (continuing with templates and values.yaml)

chart/README.md
javsalgar: Yes, that's something the CARTO team will need to handle

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,787 @@
...
+*/}}
+{{- define "carto.ldsApi.image" -}}
+{{- include "common.images.image" (dict "imageRoot" .Values.ldsApi.image "global" .Values.global) -}}
+{{- end -}}
+
+{{/*
+Return the proper Carto lds-api full name
+*/}}
+{{- define "carto.ldsApi.fullname" -}}
+{{- printf "%s-lds-api" (include "common.names.fullname" .) -}}
```
juan131, line 47: This is potentially problematic.. You should truncate to 63 characters (+trimming "-" suffixes) to avoid DNS issues when the release name

juan131: This concern applies to every "carto.XXX.fullname" template defined here.

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,787 @@
...
+      valueFrom:
+        secretKeyRef:
+          name: {{ include "carto.postgresql.secretName" .context }}
+          key: {{ include "carto.postgresql.existingsecret.key" .context  }}
+    - name: POSTGRESQL_CLIENT_POSTGRES_USER
+      value: {{ include "carto.postgresql.user" .context | quote }}
+    - name: POSTGRESQL_CLIENT_DATABASE_NAME
+      value: {{ include "carto.postgresql.databaseName" .context }}
+    {{- end }}
+{{- end -}}
```
juan131, line 641: Amazing!

chart/templates/accounts-www/secret.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if not .Values.accountsWww.existingSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "carto.accountsWww.fullname" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+    app.kubernetes.io/component: trivy
```
juan131, line 10: Wrong component?
```suggestion
    app.kubernetes.io/component: accounts-www
```

chart/templates/accounts-www/secret.yaml
```diff
@@ -0,0 +1,20 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+stringData:
+  .secret.env: |
+    # Common configuration
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonSecretConfiguration "context" $ ) | nindent 4 }}
+    # accounts-www configuration
+    {{- include "common.tplvalues.render" ( dict "value" .Values.accountsWww.secretConfiguration "context" $ ) | nindent 4 }}
```
juan131, line 19: Shouldn't this be encoded?

chart/templates/accounts-www/service.yaml
```diff
@@ -0,0 +1,51 @@
...
+    {{- end }}
+    {{- if .Values.accountsWww.service.annotations }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.accountsWww.service.annotations "context" $) | nindent 4 }}
+    {{- end }}
+spec:
+  type: {{ .Values.accountsWww.service.type }}
+  {{- if and .Values.accountsWww.service.clusterIP (eq .Values.accountsWww.service.type "ClusterIP") }}
+  clusterIP: {{ .Values.accountsWww.service.clusterIP }}
+  {{- end }}
+  {{- if ne .Values.accountsWww.service.type "ClusterIP" }}
```
juan131, line 23:
```suggestion
  {{- if and .Values.accountsWww.service.externalTrafficPolic (ne .Values.accountsWww.service.type "ClusterIP") }}
```

juan131 (changes requested):
Review for `NOTES.txt` + `_helper.tpl` + `accounts-www` manifests

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  # CARTO Self hosted configuration
+  # >>>>>>>>>>>>  TO BE CHANGED  >>>>>>>>>>>>>>>>>>>>>>>>>
+
+  # Fully qualified domain where CARTO will be accesible
+  ONPREM_DOMAIN="{{ include "carto.baseUrl" . }}"
+
+  # If you want to use Google Maps and have a API KEY set it here
+  # REACT_APP_GOOGLE_MAPS_API_KEY="CUSTOMIZED_FOR_USER_IF_NEEDED"
+
+  # Configuration onfiguration for PostgreSQL
```
juan131, line 84: Typo:
```suggestion
  # Configuration for PostgreSQL
```

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  REACT_APP_ACCOUNTS_URL="https://${ONPREM_DOMAIN}/acc/"
+  REACT_APP_WORKSPACE_API_URL="https://${ONPREM_DOMAIN}/workspace-api"
+  REACT_APP_API_BASE_URL="https://${ONPREM_DOMAIN}/api"
+  REACT_APP_PUBLIC_MAP_URL="https://${ONPREM_DOMAIN}/workspace-api/maps/public"
+  REACT_APP_AUTH0_AUDIENCE="carto-cloud-native-api"
+  REACT_APP_WORKSPACE_URL_TEMPLATE="https://{tenantDomain}"
+  REACT_APP_CUSTOM_TENANT="${ONPREM_TENANT_ID}"
+  REACT_APP_IMPORT_DATASET="carto_dw.carto-dw-{account-id}.shared"
+  REACT_APP_HUBSPOT_ID="474999"
+  REACT_APP_HUBSPOT_LIMIT_FORM_ID="cd9486fa-5766-4bac-81b9-d8c6cd029b3b"
```
juan131, line 141: Do we want to hardcode this values here?

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  ## @param accountsWww.livenessProbe.failureThreshold Failure threshold for livenessProbe
+  ## @param accountsWww.livenessProbe.successThreshold Success threshold for livenessProbe
+  ##
+  livenessProbe:
+    enabled: true
+    initialDelaySeconds: 10
+    periodSeconds: 30
+    timeoutSeconds: 5
+    successThreshold: 1
+    failureThreshold: 5
```
juan131, line 253: I'd recommend that CARTO team revisit the default seconds we're setting for every probe (readiness, liveness and startup) per component. They'll have more context about the reasonable times to wait for the micro-services to be ready/healthy/started.

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  ##
+  customStartupProbe: {}
+  ## accounts-www resource requests and limits
+  ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
+  ## @param accountsWww.resources.limits The resources limits for the accounts-www containers
+  ## @param accountsWww.resources.requests The requested resources for the accounts-www containers
+  ##
+  resources:
+    limits: {}
+    requests: {}
```
juan131, line 298: Same thing here.. I'd check with CARTO team whether we should set some default limits/requests values for the micro-services resources.

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  ## Configure Pods Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
+  ## @param accountsWww.podSecurityContext.enabled Enabled accounts-www pods' Security Context
+  ## @param accountsWww.podSecurityContext.fsGroup Set accounts-www pod's Security Context fsGroup
+  ##
+  podSecurityContext:
+    enabled: true
+    fsGroup: 1001
```
juan131, line 306: Does it make sense to use a non-root gid for the filesystem if the container is running as root?

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  ## ref: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
+  ## e.g:
+  ## initContainers:
+  ##  - name: your-image-name
+  ##    image: your-image
+  ##    imagePullPolicy: Always
+  ##    command: ['sh', '-c', 'echo "hello world"']
+  ##
+  initContainers: {}
+  ## @section router Service Parameters
```
juan131, line 1958: Should we include an Ingress Rule (+ TLS termination via TLS secrets) for this Router service?

chart/values.yaml
```diff
@@ -0,0 +1,3071 @@
...
+  ## @param workspaceMigrations.containerSecurityContext.runAsNonRoot Force the init container to run as non root
+  ## NOTE: when runAsUser is set to special value "auto", init container will try to chown the
+  ##   data folder to auto-determined user&group, using commands: `id -u`:`id -G | cut -d" " -f2`
+  ##   "auto" is especially useful for OpenShift which has scc with dynamic user ids (and 0 is not allowed)
+  ##
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 0
+    runAsNonRoot: false
+
```
juan131, line 2974: Some generic comments:

- Do we require any RBAC rule?
- NetworkPolicies?
- PDB(s) or AutoScaling?
- Metrics? If so, ServiceMonitor for Prometheus Operator?

chart/templates/NOTES.txt
```diff
@@ -0,0 +1,125 @@
...
+
+{{- else }}
+
+  echo URL  : {{ $baseUrl }}
+
+{{- end }}
+
+{{- end  }}
+
+{{ include "carto.validateValues" . }}
```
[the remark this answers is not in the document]
javsalgar: I don't think it makes sense to do it in this case as we won't know the format of the tags they plan to use

chart/templates/accounts-www/secret.yaml
javsalgar: It's stringData, so it's not necessary

chart/templates/_helpers.tpl
javsalgar: Done

chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,787 @@
...
+Add environment variables to configure database values
+*/}}
+{{- define "carto.postgresql.databaseName" -}}
+{{- ternary .Values.postgresql.postgresqlDatabase .Values.externalDatabase.database .Values.postgresql.enabled | quote -}}
+{{- end -}}
+
+{{/*
+Add environment variables to configure database values
+*/}}
+{{- define "carto.postgresql.existingsecret.key" -}}
```
juan131, line 529: This name is confusing... It suggests that it's only used for the use case of using an existing secret to configure PostgreSQL credentials.

juan131 (changes requested):
Review pending files

chart/values.yaml
javsalgar: This comes from the settings the CARTO team gives to their customers. They should be the ones deciding on this.

chart/values.yaml
javsalgar: Ok!

chart/values.yaml
javsalgar: Ok!

chart/values.yaml
javsalgar: Changed

chart/values.yaml
javsalgar: In my opinion, the ingress should substitute the router component (that's what happens in the Harbor chart). I'd leave this as something to analyze with the CARTO team but I'd leave it out of the scope of this first cut.

chart/values.yaml
javsalgar: RBAC:
  - It seems that not at the moment.
Netpol, HPA or PDB
  - We can check with the CARTO team, I'd save it for the meeting on the 30th but I don't think this is necessary for the first cut.

Metrics:
  -  I've checked their code and they don't seem to have metrics endpoints. I'll ask them in the meeting.

chart/templates/_helpers.tpl
javsalgar: Changed

juan131 (dismissed):
Approving this 1st cut. I'd keep opening the conversations that need to be checked with CARTO team. Awesome work @javsalgar

javsalgar (approved):
Shamelessly self-approving. We already performed tests using the CARTO e2e test suite and they all passed.

PR #13, opened 2022-01-13
Goss tests for accounts-www

randradas:
**Description of the change**

First goss tests for 'accounts-www' pods

**Benefits**

Now it is possible to detect regressions that could affect to 'accounts-www' component. Also there is a dir structure that will keep goss tests separated per Carto's component.

### DIFF
--- a/chart/tests/carto.yaml
+++ b/chart/tests/carto.yaml
@@ -1,0 +1,57 @@
+{
+  "phases": {
+    "package": {
+      "context": {
+        "resources": {
+          "url": "https://github.com/CartoDB/carto3-helm/archive/refs/tags/v0.0.4.tar.gz",
+          "path": "/chart"
+        }
+      },
+      "actions": [
+        {
+          "action_id": "linter-packaging",
+          "params": {
+            "kind": "HELM"
+          }
+        }
+      ]
+    },
+    "verify": {
+      "context": {
+        "resources": {
+          "url": "https://s3.amazonaws.com/bitnami-autobuild-ondemand/temporary/s3temp-bitnami-20220113161539-50692/carto-0.0.4.tgz?AWSAccessKeyId=AKIAJUJS62Y5YSCL4CIA&Expires=1642349739&Signature=d5xdKVbZ1zlkVpxujkErHPyZO2U=",
+          "path": "/tests"
+        },
+        "application": {
+          "kind": "HELM",
+          "details": {
+            "name": "carto",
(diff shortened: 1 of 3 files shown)


chart/tests/carto.yaml
```diff
@@ -0,0 +1,56 @@
...
+          "params": {
+            "kind": "HELM"
+          }
+        }
+      ]
+    },
+    "verify": {
+      "context": {
+        "resources": {
+          "url": "https://github.com/CartoDB/carto3-helm/releases/download/v0.0.4/carto3-helm.tar.gz",
```
ilbambino, line 22: I don't know how we are going to integrate these tests yet, but if the release is hardcoded, how do we plan to manage it? updating all these files on each release?

chart/tests/chart-values.yaml
```diff
@@ -0,0 +1,13 @@
+onPremDomain: "test.carto.local"
```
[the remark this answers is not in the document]
randradas: This is the value used to set the "ONPREM_DOMAIN=" var in .env during the deployment. Helm will not deploy Carto product on the environment if this value is not provided.

chart/tests/carto.yaml
randradas: Ok let me clarify, this `tar.gz` is an intermediate artifact between different phases with dependencies built-in. We just used the "release attachment" feature of GitHub to upload the `tar.gz` somewhere that can be accessed by the verification service during the `verify` phase.

Currently we are working to pass this tarball between phases, once it is done this value should not be needed.

paleloser: I think there's a misunderstanding here. `verify` phase' context is made of the following main components at the moment:

* `resources`: where the tests that will be executed are located at. This should point at a URL with the tests intended for the application being verified.
* `application`: the definition of the app to be verified.
* `runtime_parameters`: base64 encoded values to customize the app deployment for verification (current `application.values`)

Once `helm-package` is available, you'll be able to remove `application` and let the pipeline verify the result of the new action. However `resources` should be kept.

randradas: Thank you @paleloser I didn't have such detailed information about this feature, now it sounds crystal clear to me. I will include your proposed changes to the PR.

randradas: Done!

paleloser (approved):
The pipeline definition is well defined and up to date.

PR #15, opened 2022-01-20
Use the repo URL to avoid requiring to create a release

randradas:
**Description of the change**

Better use the repo URL instead of a release one.


chart/tests/carto.json
```diff
@@ -3,7 +3,7 @@
     "package": {
       "context": {
         "resources": {
-          "url": "https://github.com/CartoDB/carto3-helm/archive/refs/tags/v0.0.4.tar.gz",
```
paleloser, line 6: I'd use a commit instead, so we make sure the pipeline is reproducible and that it should work: `https://github.com/CartoDB/carto3-helm/tarball/c321361a7a237731c05572876d147cdba5190569`. WDYT?

randradas: I agree but this commit is older than the PR itself, does it make sense?

paleloser: I guess it depends of the goal of this pipeline: if it is to stay up-to-date to the code, regardless if it succeeds or not then it doesn't. But if this pipeline aims to illustrate how the engine works, with a working pipeline, then I'd fix it to a commit where we know it worked.

randradas: it's just to illustrate, done!

chart/tests/carto.yaml
```diff
@@ -0,0 +1,50 @@
+{
```
paleloser, line 1: `.json` right?

PR #16, opened 2022-01-21
Avoid put tests into '/chart' dir

randradas:
**Description of the change**

Avoid put tests into '/chart' dir so it is not included inside the redistributable

**Benefits**

Tests are not included in the helm package


randradas:
Probably I will need another PR to set the proper commit hash after the "squash&merge" as reference for the tar.

paleloser (approved):

PR #18, opened 2022-01-28
New VIB workflow 'first cut'

randradas:
**Description of the change**

This change introduces a couple of workflows that run a couple of VIB *first cut* pipelines.

**Benefits**

Automatic verification with VIB


.github/workflows/vib.yaml
```diff
@@ -0,0 +1,45 @@
+name: 'vib'
+on: # rebuild any PRs and main branch changes
+  pull_request:
+    branches:
+      - stable
+    types:
+      - assigned
```
paleloser, line 7: Why would you want to re-run the workflow if the PR is assigned? That even't doesn't mean code changes, does it?

.vib/vib-platform-verify.json
```diff
@@ -3,39 +3,33 @@
...
         }
       ]
     },
     "verify": {
       "context": {
         "resources": {
-          "url": "https://github.com/CartoDB/carto3-helm/tarball/bc22e8f8ece48277a6681ddb2dde37ca8b536ee8",
-          "path": "/tests"
+          "url": "{SHA_ARCHIVE}",
+          "path": "/chart"
```
paleloser, line 20: I'd switch this to point at the base test directory, which it looks like it's going to be `/.vib/carto/`. Then, in the actions, for example `goss`, you can use the relative path `goss/goss-accounts-www`.

.github/workflows/vib.yaml
randradas: I completely agree, I will remove it.

.github/workflows/vib.yaml
```diff
@@ -0,0 +1,44 @@
+name: 'vib'
+on: # rebuild any PRs and main branch changes
+  pull_request:
+    branches:
+      - stable
+    types:
+      - opened
+      - synchronize
+      - reopened
```
paleloser, line 9: I **think** you don't need these anymore, they are the default types for [`pull_request`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)

paleloser (approved):

PR #24, opened 2022-03-01
fix error 401 in maps & import

alberhander:
- Fix error 401 while trying to load queries from DW or connections.
- Tested imports and queries to all the supported DBs and DW.


alvarorm22 (approved):
:+1:

PR #53, opened 2022-03-16
Multiple changes

ManuelLR:
**Description of the change**

- Changing the default Service type of the router from `LoadBalancer` to `ClusterIP`.
- Rework NOTES.txt
- Rework README.md

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


README.md
```diff
@@ -46,31 +58,18 @@ Open the file `carto-values.yaml` that you have received and configure the neede
...
-  
-  In EKS:
-
-  Create a CNAME record poiting to:
-
+5. Install your deployment:
   ```bash
-  kubectl get svc carto-selfhosted-v1-router -o jsonpath='{.status.loadBalancer.ingress.*.hostname}'
+  helm install \
+    <your_own_installation_name|carto> \
```
ilbambino, line 64: I would put a default name, so all commands in the README are using the same name. If someone wants to change it to match some naming policy or whatever, they probably know how to do it. But with defaults we make the copy/paste easier

README.md
```diff
@@ -46,31 +58,18 @@ Open the file `carto-values.yaml` that you have received and configure the neede
...
+    carto-selfhosted/carto \
+    --namespace <your_namespace> \
+    -f carto-values.yaml \
+    -f carto-secrets.yaml \
+    <other_custom_files>
```
ilbambino, line 69: do we really need to add the `<other_custom_files>`??

chart/templates/router/configmap.yaml
```diff
@@ -18,10 +18,12 @@ data:
...
   ROUTER_MAPS_API_INTERNAL_URL: {{ include "carto.mapsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
-  ROUTER_SSL_AUTOGENERATE: {{ .Values.customConfigValues.routerSslAutogenerate | quote }}
+  # If required, is going to be generated by Helm and inyected as secret
+  ROUTER_SSL_AUTOGENERATE: "0"
   ROUTER_SSL_CERTIFICATE_KEY_PATH: /usr/src/certs/{{ template "carto.tlsCerts.secretKeyKey" . }}
   ROUTER_SSL_CERTIFICATE_PATH: /usr/src/certs/{{ template "carto.tlsCerts.secretCertKey" . }}
   ROUTER_WORKSPACE_API_INTERNAL_URL: {{ include "carto.workspaceApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
   ROUTER_WORKSPACE_WWW_INTERNAL_URL: {{ include "carto.workspaceWww.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
   SELFHOSTED_DOMAIN: {{ .Values.customConfigValues.selfHostedDomain | quote }}
+  ROUTER_ENABLE_HTTPS: "true"
```
ilbambino, line 28: shouldn't we put this as `false` is the LoadBalancer is going to do the TLS termination?

README.md
ManuelLR: and I don't like the default name because it is indistinguishable from the other parameters. Maybe use a BASH variable with a default value?

README.md
ManuelLR: yeah, is the way to add all the configurations like the new `Made it public` and more to come !

chart/templates/router/configmap.yaml
ManuelLR: Exact. For the moment I have left it static but when I document the ingress and certificates in the providers I will make it dynamic.

alvarorm22 (approved):
:+1:

PR #55, opened 2022-03-17
Re-estructure documentation

ManuelLR:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+  ```bash
+  ... -f <my_customization_file>.yaml
+  ```
+- Use the inline set. For example, add the following at the end of ALL the install or upgrade command:
+  ```bash
+  ... --set customConfigValues.selfHostedDomain=my.domain.com
+  ```
+
+
+## Configure the domain of your selfhosted
```
ilbambino, line 43: should we add here the TLS config? I think it would make sense

customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+Don't forget to upgrade your chart after the change.
+
+
+## Access to CARTO from outside the cluster
+By default, the `router` Service of CARTO selfhosted is configured in mode `ClusterIP` so it's only usable from inside the cluster.
+To access to it you can exec locally a [kubectl port-forward](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/) but this only make it accessible to your machine.
+
+To made it accessible from the internet (or outside the Kubernetes network) you need to properly configure the `router` Service (the easiest way) or create an Ingress.
```
ilbambino, line 62: I would skip the whole Ingress so far. Even say that so far the only supported way is through a `LoadBalancer`

customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+  tlsCerts:
+    autoGenerate: false
+    existingSecret:
+      name: "<your_own_installation_name|carto>-tls-certificate"
+      keyKey: "tls.key"
+      certKey: "tls.crt"
+  ```
+
+## Use external databases
+This package comes with an internal Postgres and Redis but it isn't the recommended way because WE DON'T CREATE BACKUPS.
```
ilbambino, line 123:
```suggestion
This package comes with an internal Postgres and Redis but it is not recommended for production. It does not have any logic for backups or any other monitoring. 
```

customizations/README.md
```diff
@@ -0,0 +1,187 @@
+# Customizations
+
+That file explain how to configure your CARTO selfhosted to meet your needs.
```
ilbambino, line 3:
```suggestion
This file explains how to configure CARTO Self-Hosted to meet your needs.
```

customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Production Ready
+
+By default, the Helm configuration with the config files provided by CARTO should work but it's not production ready.
```
ilbambino, line 7:
```suggestion
By default, the Helm configuration provided by CARTO works out of the box, but it's **not production ready**.
```

customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+In the following steps we are going to explain the steps to approximate the installation to a production ready environment.
```
ilbambino, line 8:
```suggestion
These are the steps to prepare the installation to a production ready environment.
```

customizations/README.md
ManuelLR: Pufff. Currently I only prepare it but I notice some shortcomings such as not being able to define a static IP (or at least I didn't see it for EKS).

At the moment I did not write anything and I am not going to do it as we said in the daily.

You decide if I remove it or leave it commented until we see what to do or leave it like this.

customizations/README.md
```diff
@@ -0,0 +1,187 @@
...
+So we recommend to use external databases, preferible managed database by your provider, with backups, high availability, etc.
+
+### Configure your own postgres
+CARTO selfhosted require a Postges (13+) to work.
+In that Postgres, CARTO stores some metadata and also the credentials of the external connections configured by the CARTO selfhosted users.
+
+⚠️ That Postgres has nothing to do with the connections that the user configures in the CARTO workspace since it stores the metadata of the entire CARTO selfhosted. ⚠️
+
+To proceed you need:
+- Create one secret in kubernetes with the Postgres passwords:
```
alvarorm22, line 134: I think we could add anothers comments to use the external postgres and redis without create the secret manually, using [externaldb-secret.yaml](https://github.com/CartoDB/carto-selfhosted-helm/blob/main/chart/templates/externaldb-secret.yaml) and [externalredis-secret.yaml](https://github.com/CartoDB/carto-selfhosted-helm/blob/main/chart/templates/externalredis-secret.yaml)

customizations/README.md
ManuelLR: Mmm. I put it in the section to open to the internet. Remember that the default now it's to not be opened.

customizations/README.md
ManuelLR: I only copy/pasted and little bit adapted from the old config. I think that this section should be rethinked/redo

miguelangelmorenochacon (approved):
LGTM

PR #61, opened 2022-03-21
Allow user Azure Postgresql as a backend database

miguelangelmorenochacon:
**Description of the change**

When connecting to azure postgres, the database user has the next form `user@db-name`, this is a problem when passing the `externalDatabase.user`, and `externalDatabase.adminUser` parameters, because the internal database initialization needs the user to not have the `@db-name` suffix, in order to avoid this we are adding two new (optional) values.

+ `externalDatabase.internalUser`
+ `externalDatabase.internalAdminuser`

Those values are completely optional, in case they are not provided they will get the values from `user` and `adminUser` respectively

Those values should be added alongside the other `externalDatabase` information

```yaml
externalDatabase:
  host: "postgresql-db-useast1.postgres.database.azure.com"
  user: "carto@postgresql-db-useast1"
  internalUser: "carto" # Only for Azure
  password: "asdfasdfasdfasdfa"
  adminUser: "admin_veryadmin@postgresql-db-useast1"
  internalAdminUser: 'admin_veryadmin' # Only for Azure
  adminPassword: "asdfasdfasdfasdfasdf"
  # FIXME: Why do we have values set to the 2 last parameters  if we don't have a value for the first one?
  existingSecret: ""
  existingSecretPasswordKey: "carto-password"
  existingSecretAdminPasswordKey: "admin-password"
  database: workspace_db
  port: 5432
```

**Benefits**

We can use Azure Postgresql as a backend 🚀 

**Possible drawbacks**

More complex documentation 😢 

**Applicable issues**

- https://github.com/CartoDB/cloud-native/pull/5127

**Additional information**

None


chart/templates/_helpers.tpl
```diff
@@ -322,6 +322,16 @@ Create the name of the service account to use for the workspace-api deployment
 {{- end -}}
 {{- end -}}
 
+{{/*
+If we connect to an Azure Postgres database, the internal user and the user used to connect to the database are different
```
ilbambino, line 326: I would explain this a little bit better, and even putting an example on how they look. Similar as you wrote in the readme

ilbambino (dismissed):
just small change, but lgtm

ManuelLR (approved):

PR #65, opened 2022-03-22
Increase worker nodes in VIB

FraPazGal:
Signed-off-by: FraPazGal <fdepaz@vmware.com>

**Description of the change**

Increase worker nodes in VIB

**Benefits**

Avoids insufficient resources errors.


FraPazGal:
Closed after VIB action was successfully in #67. Action's performance will be monitored for the upcoming PRs in case any performance improvements are needed.

PR #80, opened 2022-03-25
Documentation improvements

ilbambino:


customizations/README.md
```diff
@@ -1,122 +1,96 @@
-# Customizations
+# Configurations
```
ManuelLR, line 1: In Helm/Chart/K8s this is called `customization` because you adapt your package: https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing

If you would like to proceed with the non helm standards, please, rename also the folder and all the necessary paths to avoid confusion.

README.md
```diff
@@ -57,9 +60,10 @@ That files are unique per self-hosted (**couldn't be shared between multiple ins
     -f carto-secrets.yaml \
     -f customization.yaml
```
ManuelLR, line 61: Should be adapted to `client-conf.yaml` or `client-customization.yaml` if we are going to provide it.

customizations/README.md
```diff
@@ -125,7 +99,7 @@ TODO: Document and add the ability to do it
     --key=path/to/key/file
   ```
 
-- [Add the following customization](#how-to-define-customizations) lines:
+- Add the following lines to you `client-conf.yaml`:
```
ManuelLR, line 102: I have tried to put the link to how to use the custom patches in all sections so that customers who go directly to their section have a link to where it explains what the custom patches are based on and can understand it.

ManuelLR (changes requested):

customizations/README.md
ManuelLR: https://cartoteam.slack.com/archives/C02K9JU7UKF/p1648450151287389?thread_ts=1648225797.414389&cid=C02K9JU7UKF

customizations/README.md
```diff
@@ -1,122 +1,96 @@
...
 
-  router:
-    service:
-      type: LoadBalancer
-  ```
-  And add the following at the end of ALL the install or upgrade command:
   ```bash
-  ... -f customization.yaml
-  ```
-- Use the parameters as arguments. You can specify each parameter using the `--set key=value[,key=value]` argument. For example, add the following at the end of ALL the install or upgrade command:
```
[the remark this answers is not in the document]
ilbambino: but I think we should encourage the use of files and not command line. That's so easy to screw up

customizations/README.md
```diff
@@ -1,122 +1,96 @@
...
+Create a dedicated [yaml](https://yaml.org/) file `client-conf.yaml` for your configuration. For example, you could create a file with the next content:
 
-There are two ways to configure or customize the deployment:
-- [**RECOMMENDED**] Create a dedicated `customization.yaml` ([yaml official documentation](https://yaml.org/)) file. For example, you can create a file with the next content:
   ```yaml
   appConfigValues:
     selfHostedDomain: "my.domain.com"
+  ```
 
-  appSecrets:
```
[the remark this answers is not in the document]
ilbambino: I removed it because I wanted to simplify it. Eg. Having the google api key there only adds confusion, and only a few will add it

customizations/README.md
```diff
@@ -1,122 +1,96 @@
...
+- Configure a domain pointing to the exposed service.
 
-You can check an example [here](service_loadBalancer/config.yaml) (keep in mind the [general notes](#router-general-notes) before to proceed).
+### Enable and configure LoadBalancer mode
 
-<!--
-TODO: We need to talk about TLS and so on...
--->
+This is the easiest way of open your CARTO self-hosted to the world. You need to change the `router` Service type to `LoadBalancer`.
+You can find an [example](service_loadBalancer/config.yaml). But we have prepared also a few specifics for different Kubernetes versions:
```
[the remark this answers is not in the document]
ilbambino: `s/versions/flavours/` ???

customizations/README.md
```diff
@@ -1,122 +1,96 @@
...
 
 > ⚠️ CARTO self-hosted only works if the final client use HTTPS protocol. ⚠️
 
-#### Disable internal HTTPS
 <!--
+#### Disable internal HTTPS
 TODO: Document and add the ability to do it
 -->
 
-#### Configure your own certificates in CARTO
```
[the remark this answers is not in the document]
ilbambino: yes, but until it is done, it confuses more as it is an empty title

customizations/README.md
ilbambino: I think that adds more confusion than it helps

README.md
```diff
@@ -1,6 +1,6 @@
-# CARTO self-hosted [Helm chart]
+# CARTO Self Hosted [Helm chart]
 
-This repository contains the [Kubernetes Helm](https://github.com/helm/helm) chart files for CARTO Platform. Run CARTO Self Hosted in your own cloud infrastructre.
+This repository contains the [Kubernetes Helm](https://github.com/helm/helm) chart files for CARTO Platform. Run CARTO Self Hosted in your own cloud infrastructure.
```
ManuelLR, line 3: Good catch !

customizations/README.md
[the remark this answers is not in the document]
ManuelLR: We can add it in the example but commented or select another things to use in the example but I saw a lot of confusion if we don't keep a longer example

customizations/README.md
```diff
@@ -1,122 +1,107 @@
...
+
+> ⚠️ CARTO Self Hosted is not designed to be used in the path of a URL, it needs a full domain or subdomain. ⚠️
+
+To do this you need to [add the following customization](#how-to-apply-the-configurations):
 
-To do this you need to [add the following customization](#how-to-define-customizations):
 ```yaml
 appConfigValues:
   selfHostedDomain: "my.domain.com"
+appSecrets:
```
ManuelLR, line 53: you delete that from the example but add it in the section of define the domain?

maybe you missleading the section trying to fix my comments?

customizations/README.md
```diff
@@ -1,122 +1,107 @@
...
 Don't forget to upgrade your chart after the change.
 
+### Access to CARTO from outside the cluster
 
-## Access to CARTO from outside the cluster
-By default, the `router` Service of CARTO self-hosted is configured in mode `ClusterIP` so it's only usable from inside the cluster.
-To access to it you can exec locally a [kubectl port-forward](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/) but this only make it accessible to your machine.
+The entry point to the CARTO Self Hosted is through the `router` Service. By default it is configured in `ClusterIP` mode. That means it's
+only usable from inside the cluster. If you want to connect to your deployment with this mode, you need to use [kubectl port-forward]
+(<https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/>). But this only makes it accessible
```
ManuelLR, line 65: ey, that link is broken with that format

customizations/README.md
```diff
@@ -278,15 +295,15 @@ To learn how to deploy the Metrics Server, see the [metrics-server installation
...
 
-Also you can set your own preferences using the minimum and maximum values that you need in this file.
+You can edit the file to set your own scaling needs by modifying the minimum and maximum values.
 
 ### Enable static scaling
-You can set how many number of pods should have be running all time, for this, you can use the [static scale config](scale_components/static.yaml) and [configure your environment](#how-to-define-customizations)
 
-> Although we recommend the autoscaling configuration, you could choose the autoscaling feature for some components and the static configuration for the others. Remember that autoscaling override the static configuration, so if one component has both configurations, autoscaling will take precedence.
+You can set statically set the number  of pods should be running. To do it, use [static scale config](scale_components/static.yaml) adding it with `-f customizations/scale_components/static.yaml` to the the `install` or `upgrade` commands.
```
ManuelLR, line 304:
```suggestion
You can set statically set the number  of pods should be running. To do it, use [static scale config](scale_components/static.yaml) adding it with `-f customizations/scale_components/static.yaml` to the `install` or `upgrade` commands.
```

customizations/README.md
```diff
@@ -105,7 +105,7 @@ To add your own certificate you need:
   ```bash
   kubectl create secret tls \
     -n <namespace> \
-    <your_own_installation_name|carto>-tls-certificate \
+    mycarto-tls-certificate \
```
ManuelLR, line 108: And magically I remember something that @aitorch tell us.

This name is the same name generated inside the container so we should use another name to avoid collisions.

My suggestion: `mycarto-custom-tls-certificate` (That also applies to the other changes of this file)

README.md
ilbambino: after thinking about it, I don't think we need to provide it. And the doc says that you should create it. So I would leave it as it is, at least until we hear from real use

customizations/README.md
ilbambino: the problem was the new line

ManuelLR (approved):

PR #88, opened 2022-03-31
feat(chart): Add support for using other Registry for all components

miguelangelmorenochacon:
**Description of the change**

Now the value `global.imageRegistry` will be taken into account

**Benefits**

We can test images from `gcr.io/carto-artifacts`

**Possible drawbacks**

None

**Applicable issues**

- https://app.shortcut.com/cartoteam/story/219561/check-global-imageregistry

**Additional information**

None


alvarorm22 (dismissed):
LGTM

miguelangelmorenochacon:
https://app.shortcut.com/cartoteam/story/219561/check-global-imageregistry

chart/templates/_helpers.tpl
```diff
@@ -713,6 +823,28 @@ Create the name of the service account to use for the cdn-invalidator-sub deploy
...
 {{- end -}}
 
+{{/*
+Return the proper image name for cdnInvalidatorSub
+*/}}
+{{- define "carto.cdnInvalidatorSub.image" -}}
+{{- if .Values.cdnInvalidatorSub.image.name -}}
+{{- .Values.cdnInvalidatorSub.image.name -}}
+{{- else -}}
+{{- printf "%s/%s" .Values.global.imageRegistry "cdn-invalidator-sub" -}}
```
ManuelLR, line 833: Ey, that should be:
```suggestion
{{- printf "%s/%s" .Values.global.imageRegistry "consumers/cdn-invalidator-sub" -}}
```

(According with the value that you deleted from `Values.yaml`)

ManuelLR (changes requested):

chart/templates/_helpers.tpl
miguelangelmorenochacon: I overlooked it, thanks 🤗

FraPazGal:
After looking at the new functions for global values, I have a question regarding `.Values.componentName.image`. Have you considered adding `.image.registry` as additional values? 

That would help separate the registry part from the image name one, as you are now doing in the `global` section. Having separate `image.registry` and `image.name` values could also avoid having to hardcode the image name in the functions if you use `global.imageRegistry`.

This change will require further ones so it may be out of the scope of this PR, but I think it could add some more flexibility to this aspect of the chart.

PR #90, opened 2022-04-01
revert scale 0 some components if depend

alvarorm22:
**Description of the change**

There are two components that have this approach, `import-worker` and `workspace-subscriber`,  this change was introduced by us in order to only start this components if some parameters were defined in customer package config file, when this file was very young and not well structured. 

The idea was that if you execute `helm install carto` from the official helm repo (without any customer file) all pods were able to run without restarts, `import-worker` and `wrk-subscriber` needs these params, in other case they will be restarting continually, and the rests of the components do not need this params to stay in running state.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


miguelangelmorenochacon (approved):
LGTM

PR #91, opened 2022-04-01
Multiple fixes from RT review

ManuelLR:
- Fix command cdn-invalidator-sub
- Fix gcp-secrets mounted in cdn-invalidator-sub
- [x] Solve problem with `WORKSPACE_IMPORTS_PROJECTID`, `WORKSPACE_IMPORTS_KEYFILENAME` && same for  `WORKSPACE_THUMBNAILS_*` . That variables should be mounted only when a secret is passed and the projectid is defined.

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

### DIFF
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -134,10 +134,14 @@ Generate the secret def to be used in pods definitions
 {{- end -}}
 
 {{/*
-Create gcpBucketsProjectId using the gcpBucketsProjectId config or, if not defined, selfHostedGcpProjectId.
+Return true if a `appSecrets.gcpBucketsServiceAccountKey` is specified in any way
 */}}
-{{- define "carto.gcpBucketsProjectId" -}}
-{{ default .Values.cartoConfigValues.selfHostedGcpProjectId .Values.appConfigValues.gcpBucketsProjectId }}
+{{- define "carto.gcpBucketsServiceAccountKey.used" -}}
+{{- if .Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name }}
+true
+{{- else if .Values.appSecrets.gcpBucketsServiceAccountKey.value -}}
+true
+{{- end -}}
 {{- end -}}
 
 {{/*
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -137,9 +137,7 @@ Generate the secret def to be used in pods definitions
 Return true if a `appSecrets.gcpBucketsServiceAccountKey` is specified in any way
 */}}
 {{- define "carto.gcpBucketsServiceAccountKey.used" -}}
-{{- if .Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name }}
-true
-{{- else if .Values.appSecrets.gcpBucketsServiceAccountKey.value -}}
+{{- if or (.Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name) (.Values.appSecrets.gcpBucketsServiceAccountKey.value) }}
 true
 {{- end -}}
 {{- end -}}
--- a/chart/templates/import-api/configmap.yaml
+++ b/chart/templates/import-api/configmap.yaml
@@ -46,8 +46,12 @@ data:
   IMPORT_PROVIDER: {{ .Values.appConfigValues.storageProvider | quote }}
   IMPORT_BUCKET: {{ .Values.appConfigValues.importBucket | quote }}
   {{- if eq .Values.appConfigValues.storageProvider "gcp" }}
-  IMPORT_PROJECTID: {{ include "carto.gcpBucketsProjectId" . | quote }}
+    {{- if .Values.appConfigValues.gcpBucketsProjectId }}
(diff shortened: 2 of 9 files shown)


chart/templates/_helpers.tpl
```diff
@@ -134,10 +134,14 @@ Generate the secret def to be used in pods definitions
 {{- end -}}
 
 {{/*
-Create gcpBucketsProjectId using the gcpBucketsProjectId config or, if not defined, selfHostedGcpProjectId.
+Return true if a `appSecrets.gcpBucketsServiceAccountKey` is specified in any way
 */}}
-{{- define "carto.gcpBucketsProjectId" -}}
-{{ default .Values.cartoConfigValues.selfHostedGcpProjectId .Values.appConfigValues.gcpBucketsProjectId }}
+{{- define "carto.gcpBucketsServiceAccountKey.used" -}}
+{{- if .Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name }}
```
miguelangelmorenochacon, line 140: Use an OR?

```suggestion
{{- if or .Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name .Values.appSecrets.gcpBucketsServiceAccountKey.existingSecret.name }}
true
```

chart/templates/import-api/configmap.yaml
```diff
@@ -46,8 +46,12 @@ data:
   IMPORT_PROVIDER: {{ .Values.appConfigValues.storageProvider | quote }}
   IMPORT_BUCKET: {{ .Values.appConfigValues.importBucket | quote }}
   {{- if eq .Values.appConfigValues.storageProvider "gcp" }}
-  IMPORT_PROJECTID: {{ include "carto.gcpBucketsProjectId" . | quote }}
+    {{- if .Values.appConfigValues.gcpBucketsProjectId }}
+  IMPORT_PROJECTID: {{ .Values.appConfigValues.gcpBucketsProjectId | quote }}
```
miguelangelmorenochacon, line 50: Take care here because `IMPORT_PROJECTID` has a value regardless of whether    `.Values.appConfigValues.gcpBucketsProjectId` it's defined or not

https://github.com/CartoDB/carto-selfhosted-helm/blob/94209435909c1c2a1628d41aa1f2c49960d98bc4/chart/templates/_helpers.tpl#L136-L141

chart/templates/import-api/configmap.yaml
```diff
@@ -46,8 +46,12 @@ data:
...
+    {{- end }}
+    {{- if ( include "carto.gcpBucketsServiceAccountKey.used" . ) }}
```
miguelangelmorenochacon, line 52: Extra-tab?

chart/templates/workspace-api/configmap.yaml
```diff
@@ -57,10 +57,14 @@ data:
   WORKSPACE_THUMBNAILS_PUBLIC: {{ .Values.appConfigValues.workspaceThumbnailsPublic | quote }}
   WORKSPACE_THUMBNAILS_PROVIDER: {{ .Values.appConfigValues.storageProvider | quote }}
   {{- if eq .Values.appConfigValues.storageProvider "gcp" }}
-  WORKSPACE_IMPORTS_PROJECTID: {{ include "carto.gcpBucketsProjectId" . | quote }}
+    {{- if .Values.appConfigValues.gcpBucketsProjectId }}
+  WORKSPACE_IMPORTS_PROJECTID: {{ .Values.appConfigValues.gcpBucketsProjectId | quote }}
+  WORKSPACE_THUMBNAILS_PROJECTID: {{ .Values.appConfigValues.gcpBucketsProjectId | quote }}
+    {{- end }}
+    {{- if ( include "carto.gcpBucketsServiceAccountKey.used" . ) }}
```
miguelangelmorenochacon, line 64: Extra-tab?

chart/templates/_helpers.tpl
ManuelLR: I tried it, but since they are non-boolean values it can't be used like this and it fails  T.T

chart/templates/import-api/configmap.yaml
ManuelLR: I have consulted the values in production and the variable is not used so it only seems to be necessary for the bucket customs.

Any way, I didn't test the change. I'm trying to convince Moises to test that for me :see_no_evil:

chart/templates/_helpers.tpl
miguelangelmorenochacon:
```
{{- if or (eq .Values.isCar true) (eq .Values.isBus true) }}
```

chart/templates/import-api/configmap.yaml
miguelangelmorenochacon: 👍  In that case we should remove the `carto.gcpBucketsProjectId` from `_helpers.tpl`

chart/templates/import-api/configmap.yaml
miguelangelmorenochacon: I prefer to put in at the same level but not big deal 👍

chart/templates/import-api/configmap.yaml
ManuelLR: Already done

PR #93, opened 2022-04-04
Add support for Redis with TLS

miguelangelmorenochacon:
**Description of the change**

We can now connect to Redis vía TLS (and non-tls)

⚠️ By default, TLS it's **NOT** enabled

Example of `customizations.yaml`
```yaml
internalRedis:
  # With that config, we disable the internal Redis provided by the package
  enabled: false
externalRedis:
  host: "we-love-redis.redis.cache.windows.net"
  port: 6380
  password: "asdfasdfasdgasdfasdf="
  tls_enabled: true
```

**Benefits**

🥷  Hackers will not be able to see your data

**Possible drawbacks**

None, as non-TLS connections are still allowed 🕺 

**Applicable issues**

None

**Additional information**

In case you are connecting to a Redis with self-signed certificate you should add an extra parameter to pass the CA cert who signed the certificate

### DIFF
--- a/chart/templates/import-api/configmap.yaml
+++ b/chart/templates/import-api/configmap.yaml
@@ -36,9 +36,9 @@ data:
   REDIS_CACHE_PREFIX: "onprem"
   REDIS_HOST: {{ include "carto.redis.host" . }}
   REDIS_PORT: {{ include "carto.redis.port" . }}
-  REDIS_TLS_ENABLED: {{ .Values.externalRedis.tls_enabled | quote }}
-  {{- if and .Values.externalRedis.tls_enabled .Values.externalRedis.tls_ca }}
-  REDIS_TLS_CA: {{ .Values.externalRedis.tls_ca | quote }}
+  REDIS_TLS_ENABLED: {{ .Values.externalRedis.tlsEnabled | quote }}
+  {{- if and .Values.externalRedis.tlsEnabled .Values.externalRedis.tlsCA }}
+  REDIS_TLS_CA: {{ .Values.externalRedis.tlsCA | quote }}
   {{- end }}
   SELFHOSTED_DOMAIN: {{ .Values.appConfigValues.selfHostedDomain | quote }}
   SELFHOSTED_GCP_PROJECT_ID: {{ .Values.cartoConfigValues.selfHostedGcpProjectId | quote }}
--- a/chart/templates/import-api/configmap.yaml
+++ b/chart/templates/import-api/configmap.yaml
@@ -36,6 +36,10 @@ data:
   REDIS_CACHE_PREFIX: "onprem"
   REDIS_HOST: {{ include "carto.redis.host" . }}
   REDIS_PORT: {{ include "carto.redis.port" . }}
+  REDIS_TLS_ENABLED: {{ .Values.externalRedis.tls_enabled | quote }}
+  {{- if and .Values.externalRedis.tls_enabled .Values.externalRedis.tls_ca }}
+  REDIS_TLS_CA: {{ .Values.externalRedis.tls_ca | quote }}
+  {{- end }}
   SELFHOSTED_DOMAIN: {{ .Values.appConfigValues.selfHostedDomain | quote }}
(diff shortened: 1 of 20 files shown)


chart/templates/import-api/configmap.yaml
```diff
@@ -36,6 +36,10 @@ data:
   REDIS_CACHE_PREFIX: "onprem"
   REDIS_HOST: {{ include "carto.redis.host" . }}
   REDIS_PORT: {{ include "carto.redis.port" . }}
+  REDIS_TLS_ENABLED: {{ .Values.externalRedis.tls_enabled | quote }}
+  {{- if and .Values.externalRedis.tls_enabled .Values.externalRedis.tls_ca }}
```
ManuelLR, line 40: This works?
(I said because in my other PR a similar comparison didn't work)

customizations/README.md
```diff
@@ -236,6 +239,12 @@ Add the following lines to you `customizations.yaml` to connect to the external
     port: "6379"
     existingSecret: "mycarto-custom-redis-secret"
     existingSecretPasswordKey: "password"
+    tls_enabled: true
```
ManuelLR, line 242: This shouldn't be `false` to follow the default behaviour :thinking: ?

customizations/README.md
```diff
@@ -251,6 +260,12 @@ Add the following lines to you `customizations.yaml` to connect to the external
     host: <Redis IP/Hostname>
     port: "6379"
     password: ""
+    tls_enabled: true
```
ManuelLR, line 263: Same as before

customizations/README.md
miguelangelmorenochacon: Default behavior it's in `values.yaml` and already set it to `false`. Production-ready should be `true`

chart/templates/import-api/configmap.yaml
miguelangelmorenochacon: It worked when I tested 🤔  Testing it again to be secure

miguelangelmorenochacon: Yep, it's working

```yaml
externalRedis:
  host: "whatever.redis.cache.windows.net"
  port: 6380
  password: "whatever"
  tls_enabled: true
  tls_ca: |
    whatever
```

```bash
root@carto-lds-api-67b9fb885f-jvgpc:/usr/src/app# env | grep REDIS
REDIS_PASSWORD=qQzfTMn13LLnFGjbrngSRpkz5pLpbmkMZAzCaKeF0FQ=
REDIS_TLS_CA=whatever
REDIS_HOST=redis-infra-team-9568.redis.cache.windows.net
REDIS_TLS_ENABLED=true
REDIS_PORT=6380
REDIS_CACHE_PREFIX=onprem

```

chart/values.yaml
```diff
@@ -3597,19 +3597,23 @@ internalRedis:
...
+## @param externalRedis.tls_ca CA certificate in case Redis TLS cert it's selfsigned
+## @param externalRedis.existingSecret Name of an existing secret resource containing the Redis password in a 'redis-password' key
 ## @param externalRedis.existingSecretPasswordKey Key of the existing secret
 ##
 externalRedis:
   host: localhost
   port: 6379
   password: ""
+  tls_enabled: false
+  tls_ca: ""
```
alvarorm22, line 3616: why are we using `_` here? all other variables are with other format

customizations/README.md
```diff
@@ -236,6 +239,12 @@ Add the following lines to you `customizations.yaml` to connect to the external
     port: "6379"
     existingSecret: "mycarto-custom-redis-secret"
     existingSecretPasswordKey: "password"
+    tls_enabled: true
+    # If your Redis TLS Cert it's self-signed
```
alvarorm22, line 243: could we add a more explicit comment that only apply if it is a self-signed? e.g: only applies if...

chart/values.yaml
miguelangelmorenochacon: true, I'm changing it 👍

alvarorm22 (approved):
LGTM

miguelangelmorenochacon:
https://app.shortcut.com/cartoteam/story/220016/upgrade-redis-6-and-enable-tls

PR #94, opened 2022-04-04
feat(chart): scale 0 replicas if tenant info value is missing

alvarorm22:
**Description of the change**

The idea is that if you execute helm install carto from the official helm repo (without any customer file) all pods were able to run without restarts, `import-worker`, `workspace-subscriber` and `cnd-invalidator` needs these tenant config params, in other case they will be restarting continually. The rests of the components do not need this params to stay in running state.

This is a prerequisite in some marketplaces to publish the chart

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


miguelangelmorenochacon (dismissed):
What a code 😮

miguelangelmorenochacon (approved):

PR #100, opened 2022-04-11
Enable APIs to connect to Postgres with TLS

miguelangelmorenochacon:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


chart/templates/workspace-api/deployment.yaml
```diff
@@ -107,7 +107,19 @@ spec:
                 secretKeyRef:
                   name: {{ include "carto.postgresql.secretName" . }}
                   key: {{ include "carto.postgresql.secret.key" . }}
-            {{- include "carto._utils.generateSecretDefs" (dict "vars" (list "ENCRYPTION_SECRET_KEY" "WORKSPACE_THUMBNAILS_ACCESSKEYID" "WORKSPACE_THUMBNAILS_SECRETACCESSKEY" "WORKSPACE_IMPORTS_ACCESSKEYID" "WORKSPACE_IMPORTS_SECRETACCESSKEY") "context" $ ) | nindent 12 }}
+            - name: WORKSPACE_POSTGRES_SSL_MODE
```
ManuelLR, line 110: why put it here instead of in the configmap?

chart/values.yaml
```diff
@@ -3615,7 +3614,6 @@ externalRedis:
   tlsEnabled: false
   tlsCA: ""
   existingSecret: ""
-  existingSecretPasswordKey: ""
```
ManuelLR, line 3618: Question, why don't we allow to choose the key name in the client's provided secret?

chart/templates/_helpers.tpl
```diff
@@ -926,6 +926,39 @@ Add environment variables to configure database values
...
+  {{- printf "%s-%s" (include "carto.postgresql.fullname" .) "ssl-ca" -}}
+{{- else }}
+  {{- printf "%s-%s-%s" .Release.Name "externalpostgresql" "ssl-ca" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the directory where the Postgresql CA cert will  be mounted
+*/}}
+{{- define "carto.postgresql.secretMountDir" -}}
```
ManuelLR, line 943: This name is a bit confused. I was thinking all the time that the CA is going to be treated as a secret and finally it was a configmap :see_no_evil: .

Maybe we should rename that to something like... ?
```suggestion
{{- define "carto.postgresql.caMountDir" -}}
```

chart/templates/_helpers.tpl
```diff
@@ -926,6 +926,39 @@ Add environment variables to configure database values
...
+{{- print "/usr/src/certs/postgresql-ssl-ca" -}}
+{{- end -}}
+
+{{/*
+Return the filename where the Postgresql CA will be mounted
+*/}}
+{{- define "carto.postgresql.secretMountFilename" -}}
```
ManuelLR, line 950: Same as the previous, maybe rename to avoid confussions

chart/templates/_helpers.tpl
```diff
@@ -926,6 +926,39 @@ Add environment variables to configure database values
...
+{{- print "ca.crt" -}}
+{{- end -}}
+
+{{/*
+Return the absolute path where the Postgresql CA cert will be mounted
+*/}}
+{{- define "carto.postgresql.secretMountAbsolutePath" -}}
```
ManuelLR, line 957: Same :arrow_up:

chart/templates/cdn-invalidator-sub/configmap.yaml
```diff
@@ -32,4 +32,8 @@ data:
   WORKSPACE_POSTGRES_HOST: {{ include "carto.postgresql.host" . }}
   WORKSPACE_POSTGRES_PORT: {{ include "carto.postgresql.port" . }}
   WORKSPACE_POSTGRES_USER: {{ include "carto.postgresql.user" . }}
+  WORKSPACE_POSTGRES_SSL_ENABLED: {{ .Values.externalPostgresql.sslEnabled | quote }}
+  {{- if and .Values.externalPostgresql.sslEnabled .Values.externalPostgresql.sslCA  }}
```
ManuelLR, line 36: I think that we also should allow to use that config with the internalPostgresql but implement it could be a timeConsuming task... :thinking: ...

chart/templates/externalpostgresql-configmap.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if and .Values.externalPostgresql.sslEnabled .Values.externalPostgresql.sslCA }}
```
ManuelLR, line 1: That's a good idea, use another configmap. The problem is that it breaks the organization of variables.
I think that we have 2 options:
- Keep that file, move all postgres related variables of all the components to that file & mount the configmap in all components that require postgres
- Delete that file and put the certificate on each configmap of each component.

Both options implies to modify the volumes mounted on the deployments to specify the key to be mounted, as we already do with secrets

chart/templates/cdn-invalidator-sub/deployment.yaml
```diff
@@ -182,6 +187,11 @@ spec:
             items:
```
ManuelLR, line 187: Wait a second, the cdn-invalidator-sub haven't access to redis? :thinking:

chart/templates/externalredis-configmap.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if and .Values.externalRedis.tlsEnabled .Values.externalRedis.tlsCA }}
```
ManuelLR, line 1: Same opinion that with postgres

chart/templates/maps-api/deployment.yaml
```diff
@@ -167,15 +167,34 @@ spec:
           {{- if .Values.mapsApi.lifecycleHooks }}
           lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.mapsApi.lifecycleHooks "context" $) | nindent 12 }}
           {{- end }}
-          {{- if .Values.mapsApi.extraVolumeMounts }}
           volumeMounts:
+            - name: gcp-default-service-account-key
```
ManuelLR, line 171: WTF. Maps-api didn't have access to gcp secrets?????

customizations/README.md
```diff
@@ -203,6 +215,28 @@ externalPostgresql:
   ...
 ```
 
+#### Configure Postgres SSL with custom CA
+
+By default CARTO will try to connect to your Postgresql via SSL. In case you don't want to connect via SSL , which is not recommended, you can configure it via the `externalPostgresql.sslEnabled` parameters
```
ManuelLR, line 221: But the default is disabled :unstable:

customizations/README.md
```diff
@@ -212,63 +246,82 @@ In the same way as with Postgres, there are two alternatives regarding the secre
...
+  #   -----BEGIN CERTIFICATE-----
+  #   ...
+  #   -----END CERTIFICATE-----
+```
+
+> Note: One kubernetes secret is going to be created automatically during the installation process with the `externalRedis.password` that you set in previous lines.
+
+#### Configure Redis TLS
+
+By default CARTO will try to connect to your Redis via TLS. In case you don't want to connect via TLS , which is not recommended, you can configure it via the `externalRedis.tlsEnabled` parameters
```
ManuelLR, line 307: Same as before:
> But the default is disabled :unstable:

customizations/scale_components/static.yaml
```diff
@@ -4,6 +4,12 @@ accountsWww:
 importApi:
   replicaCount: 1
 
+importWorker:
```
ManuelLR, line 7: :heart:

ManuelLR (commented):
:muscle:

chart/templates/cdn-invalidator-sub/configmap.yaml
miguelangelmorenochacon: Maybe in the future, to complex to do it ATM

chart/templates/maps-api/deployment.yaml
miguelangelmorenochacon: surprise

chart/templates/workspace-api/deployment.yaml
miguelangelmorenochacon: This env var it's only required in this component

chart/templates/_helpers.tpl
miguelangelmorenochacon: We are doing this way with other similar variables

```
carto.gcpBucketsServiceAccountKey.secretMountDir
carto.google.secretMountDir
```

chart/templates/cdn-invalidator-sub/deployment.yaml
miguelangelmorenochacon: Does it need it?

chart/values.yaml
miguelangelmorenochacon: mmm I cannot remember why I removed this 🤔  Maybe because we are using the `existingSecret` approach for this?

customizations/README.md
miguelangelmorenochacon: I will change it 👍

chart/values.yaml
ManuelLR: But this is the key to look for in the Secret object in :k8s:.

If we currently don't use it, we should implement it

chart/templates/_helpers.tpl
ManuelLR: But they are "Secrets". Maybe we can leave at:
 ```suggestion
{{- define "carto.postgresql.configMapMountDir" -}}
```
or similar

chart/templates/cdn-invalidator-sub/deployment.yaml
ManuelLR: I don't know

ManuelLR: Confirmed with Tarrero that is not necessary.

ManuelLR (approved):
LGTM :muscle:

PR #128, opened 2022-05-19
Fix K8S pods where running as root user

miguelangelmorenochacon:
**Description of the change**

Kubernetes Pods where running as `root` user rather than Dockerfile defined user, `node` in this case.

**Benefits**

None

**Possible drawbacks**

None

**Applicable issues**

  - Snowflake imports were broken because of this

**Additional information**

None


aitorch (approved):
LGTM 👍🏼

PR #129, opened 2022-05-19
Add pre-hook to check minimal version for customer package

aitorch:
**Description of the change**

Added a prehook to check the customer package version vs the last breaking change version to ensure that the contents of the customer package honor the needs of the whole solution.


miguelangelmorenochacon (approved):

alvarorm22 (dismissed):
:+1:

PR #147, opened 2022-06-02
Removing subscribers from static scaling

jvillarf:
We agreed on **removing static scale** option for any subscriber, until we can test it properly.


alvarorm22 (dismissed):

ManuelLR (approved):
LGTM

PR #150, opened 2022-06-03
Allow to disable HTTPS in router

alvarorm22:
- [x] Allow to disable HTTPS in router
- [x] Add a _How to_ guide in Readme

**Description of the change**

Allow to disable HTTPS in router

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


chart/values.yaml
```diff
@@ -142,12 +142,14 @@ cartoSecrets:
       name: ""
       key: ""
 
+## @param tlsCerts.httpsRouterEnabled Enable https in router. If you disable this protocol in the Carto Router component you should configure it in a higher layer like a Load Balancer
```
FraPazGal: The change LGTM! 

Just a minor suggestion to improve the README clarity; you may want to englobe these TLS params in a specific section instead of having them under the `CARTO secrets` table. Though the TLS files are passed as a secret to the pods, with this new param you are introducing TLS config unrelated to the secret itself.

The change itself would only affect how the README tables are presented:

```suggestion
## @section TLS parameters
## Global TLS parameters
## @param tlsCerts.httpsRouterEnabled Enable https in router. If you disable this protocol in the Carto Router component you should configure it in a higher layer like a Load Balancer
```

alvarorm22: Thanks for the suggestion @FraPazGal !

miguelangelmorenochacon (approved):

PR #175, opened 2022-06-21
initial approach for use workload identity

alvarorm22:
**Description of the change**

- Workload Identity Refactor

**Benefits**

- Allow use Workload Identity

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


jvillarf (dismissed):

chart/values.yaml
```diff
@@ -158,6 +158,14 @@ tlsCerts:
     keyKey: "tls.key"
     certKey: "tls.crt"
 
+## @section Workload Identity
+## Kubernetes workload identity feature to be enabled by the client.
+workloadIdentityConfig:
+  ## @param workloadIdentityConfig.enableWorkloadIdentity Enable the Workload Identity feature in kubernetes
+  enableWorkloadIdentity: "false"
```
miguelangelmorenochacon, line 165: `true` rather than `"true"`? Is a a boolean

chart/templates/carto-google-api-service-account.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: {{ template "carto.commonGcpApi.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- if eq .Values.workloadIdentityConfig.enableWorkloadIdentity "true" }}
+    iam.gke.io/gcp-service-account: {{ .Values.workloadIdentityConfig.workloadIdentitySaEmail | quote }}
+    {{- end }}
+automountServiceAccountToken: {{ .Values.workspaceApi.serviceAccount.automountServiceAccountToken }}
```
ManuelLR, line 18: That line is using something from workspaceApi. If that is required, we should define it at a global level.

chart/templates/_helpers.tpl
```diff
@@ -204,6 +204,13 @@ Return the absolute path where the GCP Buckets Service Account Key Secret will b
 {{- printf "%s/%s" (include "carto.googleCloudStorageServiceAccountKey.secretMountDir" .) (include "carto.googleCloudStorageServiceAccountKey.secretMountFilename" .) -}}
 {{- end -}}
 
+{{/*
+Create the name of the service account to use for Carto common deployments to connect to google apis
+*/}}
+{{- define "carto.commonGcpApi.serviceAccountName" -}}
+{{- printf "%s-gcp-api" (.Chart.Name) | trunc 63 | trimSuffix "-" -}}
```
ManuelLR, line 211: We should allow to configure it in the same way that it could be configured in the other components.

chart/templates/cdn-invalidator-sub/deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
         {{- include "common.tplvalues.render" (dict "value" .Values.cdnInvalidatorSub.podLabels "context" $) | nindent 8 }}
         {{- end }}
     spec:
-      serviceAccountName: {{ template "carto.cdnInvalidatorSub.serviceAccountName" . }}
+      serviceAccountName: {{ template "carto.commonGcpApi.serviceAccountName" . }}
```
ManuelLR, line 43: I was thinking a lot regarding if that was the right way to proceed (delete the ability to customize the SA per component) and... I don't have a clear decision because if we leave the original way, we should edit all the templates so it's not easy to debug when a component uses a SA or another reading the deployment yaml but the old way allow us to attach a custom SA if required.

But... I think that, if we add the same freedom to configure the common SA as we already had with the components, we can keep this change.

aitorch (dismissed):
LGTM

chart/templates/_helpers.tpl
```diff
@@ -204,6 +204,13 @@ Return the absolute path where the GCP Buckets Service Account Key Secret will b
 {{- printf "%s/%s" (include "carto.googleCloudStorageServiceAccountKey.secretMountDir" .) (include "carto.googleCloudStorageServiceAccountKey.secretMountFilename" .) -}}
 {{- end -}}
 
+{{/*
+Create the name of the service account to use for Carto common deployments to connect to google apis
+*/}}
+{{- define "carto.commonGcpApi.serviceAccountName" -}}
```
ManuelLR, line 210: A question, why `commonGcpApi` instead of something like `commonSA`?

ManuelLR (changes requested):
Additionally:
- Review the variables `WORKSPACE_IMPORTS_KEYFILENAME` & `WORKSPACE_THUMBNAILS_KEYFILENAME` from `workspace-api` because I think that they should also be adapted in some way.
- In `values.yaml` you don't delete the config per component regarding SA.
- `lds-api` component is not using `GOOGLE_APPLICATION_CREDENTIALS`` variable but it should.
- In the `notes.txt` we should write some info when workloadIdentity is used to show the k8s SA that should be used in gcp commands

alberhander (approved):
looks good

aitorch (approved):

chart/templates/NOTES.txt
```diff
@@ -127,3 +127,18 @@ host. To configure CARTO with the URL of your service:
 
 {{ include "carto.validateValues" . }}
 {{- end }}
+
+{{- if eq .Values.commonBackendServiceAccount.enableWorkloadIdentity "true" }}
+
+3. You have enabled the workload identity feature.
+
+This is the kubernetes Service Account that you need to binding with your IAM SA in the following command or in the terraform example `https://github.com/CartoDB/carto-selfhosted/blob/master/examples/terraform/gcp/gke-autopilot.tf`
```
ManuelLR, line 135:
```suggestion
This is the kubernetes Service Account that you need to binding with your IAM SA in the following command or in the terraform example `https://github.com/CartoDB/carto-selfhosted/blob/{{- include "chart.version" . }}/examples/terraform/gcp/gke-autopilot.tf`
```

chart/templates/carto-common-service-account.yaml
```diff
@@ -0,0 +1,18 @@
...
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  name: {{ template "carto.commonSA.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- if eq .Values.commonBackendServiceAccount.enableWorkloadIdentity "true" }}
```
ManuelLR, line 15: Ey, I think that we should put the annotations always, regardless of whether or not workloadIdentity is enabled

ManuelLR (dismissed):
LGTM :tada:

PR #201, opened 2022-06-28
Sc 240571/create secret for bigquery oauth2

alvarorm22:
**Description of the change**

- Add bigqueryOauth2ClientSecret secret instead of extra var
- Add bigqueryOauth2ClientId var instead of extra var

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


chart/templates/workspace-www/configmap.yaml
```diff
@@ -12,13 +12,17 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
   {{- end }}
 data:
+  {{- if and (ne .Values.appConfigValues.bigqueryOauth2ClientId "") (ne .Values.appSecrets.bigqueryOauth2ClientSecret.value "") }}
```
ManuelLR, line 15: Ey, this is not well thought. Secret could be passed without the "value" field (remember that you can specify it from an external secret)

To simplify it, you can limit the condition to only take into account `Values.appConfigValues.bigqueryOauth2ClientId` that is going to simplify a lot and the result is going to be the same.

TIP. Check how I check if a variable is empty: https://github.com/CartoDB/carto-selfhosted-helm/pull/201/files#diff-8fd8e35b22894402e4888a546d0e531e3e36caae91af385c8eb0d42e3b80c7efR5

chart/values.yaml
```diff
@@ -19,6 +19,8 @@ appConfigValues:
   googleCloudStorageProjectId: ""
   ## @param appConfigValues.awsS3Region If the bucket is S3, the region to be used
   awsS3Region: ""
+  ## @param appConfigValues.bigqueryOauth2ClientId If the bucket is S3, the region to be used
```
ManuelLR, line 22: Please, update the docu of the variable accordingly....
```suggestion
  ## @param appConfigValues.bigqueryOauth2ClientId bla bla bla 
```

ManuelLR (approved):
LGTM.

Please remember to add documentation regarding how to add a new secret

PR #203, opened 2022-06-29
Add missing workspace vars

alvarorm22:
**Description of the change**

- Add missing env vars to workspace subscriber
- Disable lint rule - Allow start file with h2 title in Changelog

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


jvillarf (approved):
Following https://github.com/CartoDB/cloud-native/wiki/Selfhosted#how-to-add-new-secrets-in-helm , we don't need to add the secret to `_helpers.tpl` or `values.yaml`, because the secret perse already exists, so it's just a matter of passing the secret value to the right component.

PR #219, opened 2022-07-14
Update Install instructions

aitorch:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


jvillarf (changes requested):
I would remove in all places `change [your_release_name]`, because i think `[your_release_name` is enough and the other phrase is a bit misleading.

Then in step number 5, i would add a `**NOTE**` explaining that you can use any string value for `[your_release_name]` and that it's worth to write it down and you might need it in further steps (like uninstall)

PR #225, opened 2022-07-15
Add ingress options

aitorch:
**Description of the change**

According to our documentation we explain how to expose the selfhosted endpoint publicly using a google cloud load balancer, however this will create a TCP load balancer , which is ok, but definitely clients will want to have a HTTP load balancer with SSL offloading,

**Benefits**

Now you can use a L7 HTTPS loadbalancer

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

### DIFF
--- a/chart/templates/router/ingress.yaml
+++ b/chart/templates/router/ingress.yaml
@@ -1,0 +1,63 @@
+{{- if .Values.router.ingress.enabled }}
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" . }}
+kind: Ingress
+metadata:
+  name: {{ template "carto.router.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.router.ingress.annotations }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.router.ingress.annotations "context" $) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+spec:
+  {{- if and .Values.router.ingress.ingressClassName (eq "true" (include "common.ingress.supportsIngressClassname" .)) }}
+  ingressClassName: {{ .Values.router.ingress.ingressClassName | quote }}
+  {{- end }}
+  rules:
+    {{- if .Values.appConfigValues.selfHostedDomain }}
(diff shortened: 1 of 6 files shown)


customizations/ingress/gke/config.yaml
```diff
@@ -0,0 +1,50 @@
...
+      cloud.google.com/backend-config: "ingress-backend-config"
+
+## BackendConfig object is used to configure ingress features related to the backends
+## CARTO needs to extend the timeout of the connections to 600 seconds
+extraDeploy:
+  - |
+    apiVersion: cloud.google.com/v1
+    kind: BackendConfig
+    metadata:
+      name: ingress-backend-config
```
ManuelLR, line 38: Shouldn't we prefix it with carto or similar?

chart/templates/router/ingress.yaml
```diff
@@ -0,0 +1,63 @@
+{{- if .Values.router.ingress.enabled }}
```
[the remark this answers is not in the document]
aitorch: Ingress only knows about router service so, for me, yes, it should be inside router, but if we will have remove router in favour of an smart ingress or an api gateway in the future it should be out for sure.

chart/templates/router/ingress.yaml
```diff
@@ -0,0 +1,63 @@
...
+spec:
+  {{- if and .Values.router.ingress.ingressClassName (eq "true" (include "common.ingress.supportsIngressClassname" .)) }}
+  ingressClassName: {{ .Values.router.ingress.ingressClassName | quote }}
+  {{- end }}
+  rules:
+    {{- if .Values.appConfigValues.selfHostedDomain }}
+    - host: {{ .Values.appConfigValues.selfHostedDomain }}
+      http:
+        paths:
+          - path: {{ .Values.router.ingress.path }}
```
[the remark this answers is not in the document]
aitorch: There are some implementations that needs to add extra hosts and maybe routes (like ALB), So I keep it as flexible as possible

customizations/ingress/gke/config.yaml
aitorch: Agree, it is just an example but can be more complete adding a prefix, but you can see that i keep all the labels related to carto just to keep track of this object.

chart/templates/router/ingress.yaml
```diff
@@ -0,0 +1,63 @@
...
+            pathType: {{ default "ImplementationSpecific" .pathType }}
+            {{- end }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "carto.router.fullname" $) "servicePort" "http" "context" $) | nindent 14 }}
+    {{- end }}
+    {{- if .Values.router.ingress.extraRules }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.router.ingress.extraRules "context" $) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.router.ingress.extraRules }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.router.ingress.extraRules "context" $) | nindent 4 }}
+    {{- end }}
```
miguelangelmorenochacon, line 51: This is duplicated, right?

PR #228, opened 2022-07-26
Feature/optional b64 encoding

dave-planet:
**Adds a flag to toggle b64 decoding of secrets**

I have added an option to set a field `.Values.appConfigValues.b64EncodeSecrets`

It defaults to `"false"` and therefore will default to the current behavior. 

**Benefits**

Ability to leave the secret string as they are inputted, then placeholders can be used, and further template-ing can happen downstream. This is good for our use-case that we want to separate the prod and staging environments.

**Possible drawbacks**

None

**Applicable issues**

**Additional information**

I also split a composite secret.yaml into two individual files, this is because I would like all the outputted manifests to be single files. This is to easily validate them with available tools and  fits with our internal tool for deploying (expects 1 yaml per file)

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -27,6 +27,7 @@ To install, upgrade or uninstall this chart, please refer to [the root README.md
 | `appConfigValues.googleCloudStorageProjectId` | If the bucket is GCP, the ProjectId to be used                                                                         | `""`                   |
 | `appConfigValues.awsS3Region`                 | If the bucket is S3, the region to be used                                                                             | `""`                   |
 | `appConfigValues.bigqueryOauth2ClientId`      | The Client ID used in BigQuery OAuth connections using Sign in with Google instead of providing a service account key. | `""`                   |
+| `appConfigValues.b64EncodeSecrets`            | Indicate if secret values should be base64 encoded.                                                                    | `true`                 |
 
 
 ### CARTO config parameters
--- a/chart/templates/_helpers.tpl
+++ b/chart/templates/_helpers.tpl
@@ -41,6 +41,11 @@ If not using ClusterIP, or if a host or LoadBalancerIP is not defined, the value
 {{- end -}}
 
 {{/*
+Define b64EncodeSecrets variable from the outersope to be used in another context
+*/}}
+{{- $b64EncodeSecrets := .Values.appConfigValues.b64EncodeSecrets -}}
+
+{{/*
 Association between env secret and path of the secret in values.yaml
 */}}
 {{- define "carto._utils.secretAssociation" -}}
@@ -84,7 +89,11 @@ Generate secret file content for a variable if a existingSecret.name is not prov
 {{ get $mapSecrets $key }}({{ $key }})={{ $secretValue }}:{{ $secretExistingName }}:{{ $secretExistingKey }}:
 */}}
 {{- if not $secretExistingName }}
+{{- if $.b64EncodeSecrets }}
 {{ $var }}: {{ $secretValue | b64enc | quote }}  # {{ $key }}.value
+{{- else }}
+{{ $var }}: {{ $secretValue | quote }}  # {{ $key }}.value
+{{- end }}
 {{- end }}
 {{- end -}}
 
--- a/chart/templates/externalpostgresql-secret.yaml
+++ b/chart/templates/externalpostgresql-secret.yaml
@@ -13,6 +13,11 @@ metadata:
   {{- end }}
 type: Opaque
 data:
+{{- if .Values.appConfigValues.b64EncodeSecrets }}
   db-password: {{ .Values.externalPostgresql.password | b64enc | quote }}
   db-admin-password: {{ .Values.externalPostgresql.adminPassword | b64enc | quote }}
+{{- else }}
+  db-password: {{ .Values.externalPostgresql.password | quote }}
+  db-admin-password: {{ .Values.externalPostgresql.adminPassword | quote }}
+{{- end }}
 {{- end }}
--- a/chart/templates/externalredis-secret.yaml
+++ b/chart/templates/externalredis-secret.yaml
@@ -13,5 +13,9 @@ metadata:
   {{- end }}
 type: Opaque
 data:
+{{- if .Values.appConfigValues.b64EncodeSecrets }}
   redis-password: {{ .Values.externalRedis.password | b64enc | quote }}
+{{- else }}
+  redis-password: {{ .Values.externalRedis.password | quote }}
+{{- end }}
 {{- end }}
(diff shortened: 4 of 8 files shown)


aitorch:
Hi Dave, thanks for your contribution. Let me understand your proposal. 
You need a way to set the secret values as b64 encoded string instead of literal strings in your values.yaml. If that the point I think it will be better to use “existingSecret” approach, this will give you the flexibility to set the secrets at your own way and the point the secret name and the key value.

dave-planet:
Hello Aitor,

Very welcome, the changes would be beneficial for me.  I will explain what I am trying to do.

Context: 

* We have a internal tool for deploying k8s manifests. This tool takes care of prod/staging separating (through templating k8s manifests with specific placeholders),  encryption of secrets and deploying the k8s manifests into the cluster. 

* I want to use this chart, in a pipeline that would first do `helm template` and then `internaltool deploy --env staging`, but not b64 encode the literal string for the secret values. (The tool will do that when evaluating the templates, so secrets will be encoded when they hit the cluster)

example:

I would like to feed the chart specific placeholder values, though the customization.yaml 

```
appConfigValues:
  googleCloudStorageProjectId: "{{ project_id }}"
 
...
 
externalPostgresql:
  password: "{{ tf_secrets.postgres_workspace_admin_credentials.password|b64encode }}"
  adminPassword: "{{ tf_secrets.postgres_admin_credentials.password|b64encode }}"
```
And have the helm template output such that these values are not b64 encoded. 

Using the  customization.yaml values above, I want `helm template ...` to output a file `externalpostgresql/secret.yaml` with the placeholders not encoding. This would be what I want
```
---
# Source: carto/templates/externalpostgresql-secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: planet-carto-externalpostgresql
  labels:
    app.kubernetes.io/name: carto
    helm.sh/chart: carto-1.38.4
    app.kubernetes.io/instance: planet-carto
    app.kubernetes.io/managed-by: Helm
  namespace: "{{ environment }}"
type: Opaque
data:
  db-password: "{{ tf_secrets.postgres_workspace_admin_credentials.password|b64encode }}"
  db-admin-password: "{{ tf_secrets.postgres_admin_credentials.password|b64encode }}"
```

The changes I submitted should do this, and not change the default behavior in anyway. The only drawback I see is that if you need to add any new secret definitions in the future, they should have the if/else pattern used.

```
{{- if .Values.appConfigValues.b64EncodeSecrets }}
  foo: {{ .Values.foo | b64enc | quote }}
  bar: {{ .Values.bar | b64enc | quote }}
{{- else }}
  foo {{ .Values.foo | quote }}
  bar: {{ .Values.bar | quote }}
{{- end }}
```

dave-planet:
Let me know if you have any questions. Other wise have a good one

aitorch:
Hi Dave,

Now I understand your use case, the case is if we merge this PR we will have a case where the generated templates will be not valid, in the sense that it will be possible to create a secret with a `data` that is not base64 encoded. From our point of view it will cause more problems than the ones it solves.

I think you have this alternatives.

1. Create secrets manifest  outside the helm and use the existing secret key to map the secret with the helm result. It should work seamly
2. Same of 1 but create secrets in `extraDeploy` value, this will create the secrets template as your wish.
3. Use the template tool before running helm (but one time per environment), i think it is the worse option.

Please let me know if it works for you.

dave-planet:
Hello and thanks for getting back to me on it,

Regarding the case you described, that should only be the case if a user **explicitly** sets the `.Values.appConfigValues.b64EncodeSecrets` value to `true`, in their specific `customization.yaml`.

That is because the default is set to `false` [here](https://github.com/CartoDB/carto-selfhosted-helm/pull/228/files#diff-a9d07052701ab3b718c82ac0fe74b965a8260b06bb5906362081718b0ee593b4R27)

For example consider the case that you have the following three `customization.yaml` files to feed into helm. Only the third will affect the way that secrets are created and create the situation of un-encoded data (and hence a broken secret)

```
---
# customization.yaml

appConfigValues:
  selfHostedDomain: "yolo"
  storageProvider: "gcp"
  httpCacheEnabled: "true" 
  importBucket: "carto-import-bucket"
  workspaceImportsBucket: "carto-client-bucket"
  workspaceImportsPublic: true
  workspaceThumbnailsBucket: "carto-thumbnails-bucket"
  workspaceThumbnailsPublic: false
```

```
---
# customization.yaml

appConfigValues:
  selfHostedDomain: "yolo"
  storageProvider: "gcp"
  httpCacheEnabled: "true" 
  importBucket: "carto-import-bucket"
  workspaceImportsBucket: "carto-client-bucket"
  workspaceImportsPublic: true
  workspaceThumbnailsBucket: "carto-thumbnails-bucket"
  workspaceThumbnailsPublic: false
  b64EncodeSecrets: false
```

```
---
# customization.yaml

appConfigValues:
  selfHostedDomain: "yolo"
  storageProvider: "gcp"
  httpCacheEnabled: "true" 
  importBucket: "carto-import-bucket"
  workspaceImportsBucket: "carto-client-bucket"
  workspaceImportsPublic: true
  workspaceThumbnailsBucket: "carto-thumbnails-bucket"
  workspaceThumbnailsPublic: false
  b64EncodeSecrets: true
```

What do you think about accepting this PR with this context? If it helps, I can add a better description/warning into the `values.yaml` to be rendered here and circumvent any problems users may experiance if they set the value to `true`.

dave-planet:
Hi Aitor,

Any updates on this? If not no worries.

Regards,
Dave

dave-planet:
Hi Aitor,

Any updates here?

After a long wait, our companies deal with Carto is signed and we will be deploying the platform based on this helm chart.

I kindly ask if you would consider merging these changes now. It makes arguably more sense for these changes to be upstream with you and the community, instead of us maintaining a fork. I am happy putting in PR's to this repo if any future changes need to be made to accommodate the feature.

As mentioned in the above comment, the changes add a sane default which will have no effect on any chart output. If the default is over-ridden, then the output will be more compatible with our deployment tool (other users may also benefit).

Let me know and have a good one!

Dave

alroba:
Hi @dave-planet,

Thank you for the reminder :)

Our Engineering Team has been reviewing your request and they believe this use case don't fit our developments at the moment.

Nevertheless, as Aitor commented, you can achieve the same behavior by creating the secrets outside the chart and referencing them from the `customizations.yaml`. You can generate the name of the secret depending on the tool environment.

We are sorry for the inconveniences and really hope the workaround is useful for you.

Best,

dave-planet:
Hello Alvaro,

That is ok, thanks regardless for reviewing it.

All the best,
Dave

PR #255, opened 2022-08-31
Add-neg-config-to-ingress-example

aitorch:
**Description of the change**

Added comments to the ingress/gke example to include 
`cloud.google.com/neg: '{"ingress": true}'`

**Benefits**

Cover the case were container native load balancing is not default

### DIFF
--- a/customizations/README.md
+++ b/customizations/README.md
@@ -142,6 +142,9 @@ But this only makes it accessible to your machine.
 
   Within this option you could either use your own TLS certificates, or GCP SSL Managed Certificates.
 
+  > :warning: if you are running a GKE cluster 1.17.6-gke.7 version or lower, please check [Cluster IP configuration](#troubleshooting)
+
+
   **Useful links**
 
   - [google-managed-certs](https://cloud.google.com/load-balancing/docs/ssl-certificates/google-managed-certs#caa)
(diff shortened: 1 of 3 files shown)


jvillarf (changes requested):
For my mindset is more complex to think about `NOT` meeting requirements instead of meeting them, so this is what i propose:

```
# https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing 
#
## To use ClusterIP the service needs to point to a NEG
## If your installation matches with the configuration below, *** uncomment the neg annotation below ***:
##
##  * GKE cluster running 1.17.6-gke.7 version or lower
##  * Not using a VPC-native cluster
##  * Using a Shared VPC
##  * Using GKE Network Policy
##
#cloud.google.com/neg: '{"ingress": true}'
```
- For shared VPC and GKE Network policy, i'm not sure if it's an `OR` , or an `AND`, so if you accept the changes, please update it accordingly.

Additionally, in the `ingress`section i would add

```
> :warning: if you are running a GKE cluster 1.17.6-gke.7 version or lower, please check [Cluster IP configuration](link here pointint to the troubleshooting section)
```

a

jvillarf (approved):

PR #273, opened 2022-10-13
Increase min containers for maps and workspace API from 1 to 2

miguelangelmorenochacon:
**Description of the change**

Increase min replicas for `workspace-api` and `maps-api` from 1 to 2

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


jvillarf (changes requested):
This is wrong, what we need is:
- rename `static.yaml` to `development.yaml`
- Modify the `chart` itself, so the `replicaCount=2` is the default value for both `maps-api` and `workspace-api` internally. This file is just a "recommendation" and what we want is that anyone installing this took the `replicaCount = 2` by default, so it must be changed at char level.

ilbambino (approved):

miguelangelmorenochacon:
https://app.shortcut.com/cartoteam/story/264056

alvarorm22 (commented):
You have to change this file reference in vib execution https://github.com/CartoDB/carto-selfhosted-helm/blob/main/.github/workflows/vib.yaml#L200

alvarorm22 (approved):

jvillarf (approved):

PR #294, opened 2022-11-30
Avoid the installation of workflows-engine for E2E

ManuelLR:
**Description of the change**

In the next release, we're going to add a new private dependency not required for the e2e-test so we should avoid the installation on VIB.

**Benefits**

Avoid grant more permission to a private package not required for e2e

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


ilbambino (approved):
🤞

PR #303, opened 2023-01-10
Allow specifying a global imageRegistry and imageTag

miguelangelmorenochacon:
**Description of the change**

Allow to specify a global `imageTag` for all the deployments docker images

Example
```yaml
global:
  imageTag: "latest"
```

**Benefits**

We don't need to specify the image tag for every single component

**Possible drawbacks**

None known.

**Applicable issues**

**Additional information**


alberhander (approved):
nice!

PR #316, opened 2023-03-16
Update doc resources with routerMetrics component

alvarorm22:
**Description of the change**

- [x] Update doc resources with router-metrics component

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


jvillarf (changes requested):
I've just realized that we are not setting the `jvm.options.heap_size` in logstash: https://www.elastic.co/guide/en/logstash/current/jvm-settings.html#heap-size (it means handling an extra file `jvm.options`), where we should follow one of the best practices:

> Do not increase the heap size past the amount of physical memory. Some memory must be left to run the OS and other processes. As a general guideline for most installations, don’t exceed 50-75% of physical memory. The more memory you have, the higher percentage you can use.

For me, the `HEAP size` should be always, let's say arond `70%` of the max physical memory (limits.memory) and for "big installations" requiring more memory, then we just increase the `requests` and `limits` parameters. 

And at this point, may be you are right and both `requests` and `limits` memory figures should remain the same.

alvarorm22:
https://app.shortcut.com/cartoteam/story/303892/update-selfhosted-doc-resources-with-latest-components

jvillarf (approved):

PR #339, opened 2023-04-12
add terminationGracePeriodSeconds

alberhander:
- Add `terminationGracePeriodSeconds` to containers

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -418,6 +418,10 @@ accountsWww:
     capabilities:
       drop:
         - all
+  ## @param accountsWww.terminationGracePeriodSeconds Time to wait before force killing the container
+  ##
+  terminationGracePeriodSeconds: 60
+
   ## @param accountsWww.existingConfigMap The name of an existing ConfigMap with your custom configuration for accounts-www
   ##
   existingConfigMap: ""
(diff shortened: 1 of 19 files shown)


chart/values.yaml
```diff
@@ -993,6 +1000,9 @@ importWorker:
     capabilities:
       drop:
         - all
+  ## @param importWorker.terminationGracePeriodSeconds Time to wait before force killing the container
+  ##
+  terminationGracePeriodSeconds: 300
```
jvillarf, line 1005: Shouldn't it be `600` for Import worker?

jvillarf (approved):
I don't know if i'm checking it in the right place, but checking the timeouts from https://github.com/CartoDB/cloud-native/blob/main/terraform/docs/project/carto-tnt-gcp-us-east1-1.md :
- `import-worker` : `600s`
- `workspace-api`: `30s`

`Worspace-api` is quite small, but probably there is a reason behind it. In fact, the table is wrong / has a typo because it is properly set to `300` in https://github.com/CartoDB/cloud-native/blob/main/terraform/_modules/gcp-tenant/cloud-run-workspace.tf#L121.

And the import-worker one is https://github.com/CartoDB/cloud-native/blob/main/terraform/_modules/gcp-tenant/cloud-run-import.tf#L115

So never mind, i'm probably mixing `timeouts` here.

miguelangelmorenochacon (approved):

PR #348, opened 2023-05-09
Add troubleshooting related to nginx 413 error code

alvarorm22:
**Description of the change**

Doc: Add troubleshooting related to the Nginx `413 Request Entity Too Large` code error

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

### DIFF
--- a/customizations/README.md
+++ b/customizations/README.md
@@ -1427,13 +1427,13 @@ Now you can run the upgrade operation again
 
   In Nginx, you have to add this config line to your location service:
 
-    ```bash
-    client_max_body_size 10M;
-    ```
+  ```bash
+  client_max_body_size 10M;
+  ```
 
   If you have an Ingress Nginx, you have to add the following annotation:
 
-    ```yaml
-      annotations:
-        nginx.ingress.kubernetes.io/proxy-body-size: "10m"
-    ```
+  ```yaml
+    annotations:
+      nginx.ingress.kubernetes.io/proxy-body-size: "10m"
+  ```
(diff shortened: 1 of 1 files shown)


customizations/README.md
```diff
@@ -1414,3 +1416,24 @@ REVISION	UPDATED                 	STATUS         	CHART             	APP VERSION
 29      	Tue Oct  4 10:58:22 2022	deployed       	carto-1.42.10-beta	2022.9.28  	Rollback to 27
 ``` 
 Now you can run the upgrade operation again
+
+### 413 Request Entity Too Large
+
+  You are trying to make a POST request to the SQL Api with a large payload. Please see the following considerations:
```
jvillarf, line 1422:
```suggestion
  You are trying to make a POST request, probably to the `SQL Api`, with a **large payload**. Please take into account the considerations below:
```

customizations/README.md
```diff
@@ -1414,3 +1416,24 @@ REVISION	UPDATED                 	STATUS         	CHART             	APP VERSION
...
+
+  - We support a payload up to 10Mb
+
+  - If your payload is lower than 10Mb, probably the error code is returned by a service in a higher layer than the Carto Selfhosted environment. Please upload your service configuration to be able to manage higher requests. E.g:
```
jvillarf, line 1426:
```suggestion
  - If your payload is lower than 10Mb, probably the error code is returned by a service in a higher layer than the CARTO Selfhosted environment. Please **update** your service configuration to be able to manage requests with a higher payload. E.g:
```

customizations/README.md
```diff
@@ -1414,3 +1416,24 @@ REVISION	UPDATED                 	STATUS         	CHART             	APP VERSION
...
+
+  In Nginx, you have to add this config line to your location service:
```
jvillarf, line 1428:
```suggestion
  In Nginx, you have to add this config line to your location configuration:
```

jvillarf: I think we should explain this better, the use case of an `nginx ingress` is clear, but i'm not sure if the users will understand the other use case "`nginx`". Probably we need to elaborate this better

jvillarf (changes requested):

PR #358, opened 2023-05-25
bigquery oauth variables for import-worker

alberhander:
- Add missing variables to import-worker component in order to enable imports with Bigquery Oauth connections.


miguelangelmorenochacon (approved):
👍

PR #361, opened 2023-05-30
Fix typos in download-customer-package code

alroba:
**Description of the change**

- Discovered a typo that created a file called `1` every time we execute the tool.
  - Fixed changing `2&>1` to `2>&1`.
- Added the required ":" to the options with parameters in `getopt`.

**Benefits**

Avoid creating an extra file when executing the tool.

`getopt` will work properly with the long options.

**Possible drawbacks**

There shouldn't be any

**Additional information**

While debugging this, we found out the following:
- We create a file for the customer service account that we might be interested in removing at the end of the script.
- This script won't work on MAC because it uses a different `getopt`.
  - It will work if we install `gnu-getopt` over the default `getopt`.
  - We can create a warning in the documentation (even though we specifically require a Linux machine to execute this script).

### DIFF
--- a/tools/carto-download-customer-package.sh
+++ b/tools/carto-download-customer-package.sh
@@ -16,7 +16,7 @@ function check_deps()
 {
   for DEP in ${DEPENDENCIES}; do
     # shellcheck disable=SC2261,SC2210
-    command -v "${DEP}" 2&>1 > /dev/null || \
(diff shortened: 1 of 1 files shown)


alroba:
The `getopt` vs `gnu-getopt` in Mac already affected a customer (Indosat): https://carto.zendesk.com/agent/tickets/86771

PR #378, opened 2023-07-10
[324925] Fix download customer package script to be compatible with MacOS

pnazar-carto:
## Shortcut

- Story: https://app.shortcut.com/cartoteam/story/324925/fix-download-customer-package-script-to-be-compatible-with-macos
- Autolink: [sc-324925]

## Description of the change

In order for `carto-download-customer-package.sh` to support `getopt` long-options in macOS it would require these actions from customer-side: `gnu-getopt` installation + binary linking.
As the above is invasive and error prone, the proposal is to change the usage of `getopt` binary by `getopts` bash builtin function which is native to GNU-Bash packed in both Linux and Darwin systems.

## Benefits 👍 

1. NO customer-side actions required (ge: installing/linking new OS packages)
2. Faster to implement

## Possible drawbacks ⚠️

This change will completely remove the long-options usage from `carto-download-customer-package.sh` and short-options will be the only allowed method to pass arguments to the script.

The new usage with short-options-only would be:

```
   Usage: carto-download-customer-package.sh [-d dir_path] [-s <k8s|docker>]

   optional arguments:
     -d    Folder path where both <carto-values.yaml> and <carto-secrets.yaml> are located (k8s)
         or where both <customer.env> and <key.json> are located (docker). 
           Default is current directory.
     -h    Show this help message and exit
     -s    Selfhosted-mode for the customer package: k8s or docker.
           Default is k8s.
```

## Applicable issues

  - fixes #324925

**Additional information**

## Tests

```
➜  carto-selfhosted-helm ✗ uname -s -m
Darwin arm64

➜  carto-selfhosted-helm ✗ bash --version | head -1
GNU bash, version 5.2.15(1)-release (aarch64-apple-darwin22.1.0)

➜  carto-selfhosted-helm ✗ tools/carto-download-customer-package.sh -d ~/tmp/package/pnazar-test-1 -s docker

  #========================================================
  FILE_DIR                     = ~/tmp/package/pnazar-test-1
  SELFHOSTED_MODE              = docker
  #========================================================

```

### DIFF
--- a/tools/carto-download-customer-package.sh
+++ b/tools/carto-download-customer-package.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 
 ##########################################
-# Requirements: yq gsutil
+# Requirements: yq jq gsutil gcloud
 ##########################################
 DEPENDENCIES="yq jq gsutil gcloud"
 SELFHOSTED_MODE="k8s"
-FILE_DIR=""
+FILE_DIR="."
 CARTO_SERVICE_ACCOUNT_FILE="./carto-service-account.json"
 CLIENT_STORAGE_BUCKET=""
 CUSTOMER_PACKAGE_NAME_PREFIX="carto-selfhosted-${SELFHOSTED_MODE}-customer-package"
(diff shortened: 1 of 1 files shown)


miguelangelmorenochacon (approved):
I haven't tested. Looks good ✅

PR #393, opened 2023-08-08
:elephant: Add support for non-default Postgresql's postgres database and non-default port for self-hosted deployments

miguelangelmorenochacon:
Shortcut: https://app.shortcut.com/cartoteam/story/338514/allow-nondefault-root-database-and-admin-user-in-selfhosted-init-db-script

The default Postgresql's `postgres` database is hard-coded in the init-db (`workspace-migrator`) script, and so does the default port (5432). We are adding support for a custom root  database and port for this script


alvarorm22 (approved):
awesome

PR #440, opened 2024-01-08
feat: init with carto replicated beta

alvarorm22:
**Description of the change**

- [x] Add carto replicated
- [x] Initial customization values and configuration options
- [x] Work with preflights and support-bundle 
- [x] Force to use external postgres in replicated flavour 
- [x] Allow to use custom Service Account with Workload Identity 

Tests:
- [x] Test manual Helm installation without Kots
- [x] Test Kots installation with custom Postgres
- [x] Test Kots installation with workload Identity
- [x] Test Kots installation with Default SA 
- [x] Test Kots installation with Redis as a container
- [x] Test Kots installation with custom Redis   

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


chart/Chart.yaml
```diff
@@ -16,6 +16,10 @@ dependencies:
     alias: internalRedis
     repository: https://charts.bitnami.com/bitnami
     version: 16.x.x
+  - name: replicated
+    repository: oci://registry.replicated.com/library
+    version: 1.0.0-beta.13
+    condition: replicated.enabled
```
ManuelLR, line 22: What is going to happen if we don't set that value on `values.yaml`?
We should avoid it on regular installacion, right?

chart/templates/preflight.yaml
```diff
@@ -0,0 +1,127 @@
+apiVersion: v1
```
ManuelLR, line 1: Taking into account that we are not 100% sure that this is going to be the final implementation and could leave leftover on our clients... What do you think if we wrap all that content with a condition like in https://github.com/CartoDB/carto-selfhosted-helm/blob/d506900ada50e7685167d752f514c5da95476be3/chart/templates/router/ingress.yaml#L1 ? 

For example, we can use the same "replicated.enabled" condition like in the chart.

chart/templates/preflight.yaml
```diff
@@ -0,0 +1,127 @@
...
+            collectorName: workspace-db
+            outcomes:
+              - fail:
+                  when: connected == false
+                  message: Cannot connect to PostgreSQL server
+              - pass:
+                  when: connected == true
+                  message: The PostgreSQL server is available
+        - postgres:
+            checkName: PostgreSQL must be v11.x or later
```
ManuelLR, line 58: We should check that the version is, at least, 14 as we said in our public documentation: https://docs.carto.com/carto-self-hosted/key-concepts/deployment-requirements#external-database

chart/templates/preflight.yaml
```diff
@@ -0,0 +1,127 @@
...
+                  message: Cannot connect to PostgreSQL server
+              - fail:
+                  when: version < 11.x
+                  message: The PostgreSQL server must be at least version 11
+              - pass:
+                  message: The PostgreSQL verion checks out
+        - clusterVersion:
+            outcomes:
+              - fail:
+                  when: "< 1.20.0"
```
ManuelLR, line 72: In our old documentation we had as minimum the `1.12` but looks too old. Whats your opinion here @moicalcob ?

(This is not a blocker for the PR but is something that we should decide in some way)

chart/templates/preflight.yaml
```diff
@@ -0,0 +1,127 @@
...
+              - pass:
+                  when: "== digitalocean"
+                  message: DigitalOcean is a supported distribution.
+              - warn:
+                  message: Unable to determine the distribution of Kubernetes.
+        - nodeResources:
+            checkName: Total CPU Cores
+            outcomes:
+              - fail:
+                  when: "sum(cpuCapacity) < 2"
```
ManuelLR, line 121: Here we should define 5 as our doc say and recommend as 6 for example.
Based on our public doc https://docs.carto.com/carto-self-hosted/key-concepts/deployment-requirements#orchestrated-deployment-kubernetes

chart/templates/support-bundle.yaml
```diff
@@ -0,0 +1,217 @@
+apiVersion: v1
```
ManuelLR, line 1: All comments of the previous files should be checked also on that file.

chart/templates/support-bundle.yaml
```diff
@@ -0,0 +1,217 @@
...
+stringData:
+    # To use it: kubectl support-bundle --load-cluster-specs --selector=carto.com/kind=support-bundle
+  support-bundle-spec: |
+    apiVersion: troubleshoot.sh/v1beta2
+    kind: SupportBundle
+    metadata:
+      name: {{ .Release.Name }}-support-bundle
+    spec:
+      collectors:
+        - clusterInfo: {}
```
ManuelLR, line 17: That could be filtered by namespace or doesn't sound logical? :thinking:

chart/templates/preflight.yaml
```diff
@@ -0,0 +1,127 @@
...
+  labels:
+    troubleshoot.sh/kind: preflight
+  name: {{ .Release.Name }}-preflight-config
+  namespace: {{ .Release.Namespace | quote }}
+stringData:
+  preflight.yaml: |
+    apiVersion: troubleshoot.sh/v1beta2
+    kind: Preflight
+    metadata:
+      name: preflight
```
ManuelLR, line 13: I think that here we should also add the namespace or could I be wrong? :thinking:

manifests/kots-app.yaml
```diff
@@ -0,0 +1,23 @@
+---
+apiVersion: kots.io/v1beta1
+kind: Application
+metadata:
+  name: Carto
+spec:
+  title: CARTO
+  icon: https://upload.wikimedia.org/wikipedia/commons/thumb/1/13/CARTO-logo.svg/1280px-CARTO-logo.svg.png
+  allowRollback: false
```
ManuelLR, line 9: Here is the line ! ! 
I know that we had the option to enable/disable rollback but I didn't remember it

chart/Chart.yaml
alvarorm22: you are right, we have to disable in the `values.yaml` file

chart/templates/preflight.yaml
alvarorm22: perfect, I'll do that, the same for support-bundle

chart/templates/support-bundle.yaml
alvarorm22: the `clusterInfo` collector can't be filtered by namespace, it only show the kubernetes version and a few more parameters

ManuelLR (dismissed):
LGTM

chart/templates/preflight.yaml
moicalcob: I've been checking it and it seems that the `1.12` version is more than 5 years old. IMHO we should update it and ask at least for the `1.25`version 🤔

alvarorm22: Ok we can put min 1.25 and recommended >= 1.26

ManuelLR: We should leave the code as it is now but I should add to the AD the official upgrade of the version and the doc changes

chart/templates/preflight.yaml
```diff
@@ -79,20 +79,20 @@ stringData:
             collectorName: redis
             outcomes:
               - fail:
-                when: isConnected == false
-                message: Cannot connect to the Redis instance
+                  when: isConnected == false
```
ManuelLR, line 82: :see_no_evil:

chart/templates/preflight.yaml
ManuelLR: `1.25` looks a decent version by the time customers start using this. in GKE: https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule-for-release-channels

ManuelLR (dismissed):
Let's go ! !

chart/templates/preflight.yaml
```diff
@@ -89,11 +89,11 @@ stringData:
               - fail:
                   when: "< 1.25.0"
                   message: The application requires Kubernetes 1.25.0 or later, and recommends 1.26.0 or later.
-                  uri: https://www.kubernetes.io
+                  uri: hhttps://kubernetes.io/releases
```
ManuelLR, line 92: Ey, that looks a typo

ManuelLR (approved):

PR #453, opened 2024-01-31
Add tenant requirements checker component

alvarorm22:
- Add workspace-db preflights with the `tenant-requirements-checker` component

### DIFF
--- a/chart/templates/_commonChecks.tpl
+++ b/chart/templates/_commonChecks.tpl
@@ -45,10 +45,11 @@ Return common collectors for preflights and support-bundle
 Return common analyzers for preflights and support-bundle
 */}}
 {{- define "carto.replicated.commonChecks.analyzers" }}
+  {{- range list "Check_database_connection" "Check_database_encoding" "Check_user_has_right_permissions" "Check_database_version" }}
   - jsonCompare:
-      checkName: Check Database Connection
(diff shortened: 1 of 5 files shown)


chart/templates/_commonChecks.tpl
```diff
@@ -32,49 +25,41 @@ Return common collectors for preflights and support-bundle
...
-        - fail:
-            when: connected == false
-            message: Cannot connect to PostgreSQL server
-        - pass:
-            when: connected == true
-            message: The PostgreSQL server is available
-  - postgres:
-      checkName: PostgreSQL must be v14.x or later
-      collectorName: workspace-db
+  {{- range list "Check_database_connection" "Check_database_encoding" "Check_user_has_right_permissions" "Check_database_version" }}
```
ManuelLR, line 48: What do you think of define it as a template and call for each checker?
Right now, looks so impersonal. Do you think that is going to be solved with the option to extract information from the container message?

alvarorm22: Yes, if you refer to the message that we are going to see in each check, we will see the same message that the tenant-requirements-checker container is logging, e.g:

`Postgres version is 13.12, but it should be at least 13.2`

`User permissions are correct`

`Postgres connection successful!`


and so on...

ManuelLR (approved):
LGTM

PR #455, opened 2024-02-01
Add Kots config for high availability settings

moicalcob:
**Description of the change**
Add Kots config for hight availability settings. From now on we'll be able to define the min/max number of instances for the pods running in k8s.


ManuelLR (approved):
LGTM

PR #463, opened 2024-02-06
Activate data synchronizing between SH and Accounts

moicalcob:
**Description of the change**
With this change, we'll enable the data synchronizing between a SH installation and our Accounts part. Each time that the workspace subscriber is started, we'll check if there is any pending change to be done.


alvarorm22 (approved):

VictorVelarde:
Thx @moicalcob let's keep an eye on this in future SH installations!

PR #470, opened 2024-02-14
set buckets default values only for dev license type

alvarorm22:
**Description of the change**

- [x] We should only set the default value on google cloud storage if is a dev license.

In other cases, we should leave unsetted


moicalcob (approved):

moicalcob (approved):
Thanks so much! 😄

PR #480, opened 2024-02-21
Add Kots config for SSO

moicalcob:


manifests/kots-config.yaml
```diff
@@ -584,6 +584,22 @@ spec:
           help_text: |-
             Specify if CARTO Self-hosted should check if the proxy certificate is valid or not.
 
+        ## SSO
```
ManuelLR, line 587: I think that in some moment, someone should to re-order the config in some way 🙈

anyway, is not important at all while they preserve the IDs 🙂

ManuelLR (approved):
LGTM

PR #484, opened 2024-02-26
Improve Kots config for imports & exports

moicalcob:


manifests/kots-config.yaml
```diff
@@ -720,58 +719,68 @@ spec:
           type: bool
           default: "0"
 
-        ## Imports / Exports
-        - name: datawarehousesExportsImportsEnabled
-          title: Imports & Exports in Datawarehouses
+        # Bigquery exports
+        - name: datawarehousesExportsBigQueryEnabled
+          title: Enable exports from BigQuery
```
ManuelLR, line 724: we should also specify CDW in some way? :thinking:

manifests/kots-config.yaml
```diff
@@ -720,58 +719,68 @@ spec:
...
-        - name: datawarehousesExportsImports
-          type: heading
-          title: Imports & Exports in Datawarehouses
-          when: '{{repl (ConfigOptionEquals "datawarehousesExportsImportsEnabled" "1") }}'
+            Enable exporting data from BigQuery connections.
         - name: storageBucketsExportGcp
-          title: GCP exports bucket name
-          help_text: The bucket name to store exported data from BigQuery
+          title: BigQuery exports bucket name
+          help_text: The bucket name to store exported data from BigQuery. If you need further info, please check [docs.carto.com](https://docs.carto.com/~/changes/1Lia4Pv0PjH33F9R0CyP/carto-self-hosted/guides/configure-your-own-buckets#configure-data-export-bucket-for-bigquery).
```
ManuelLR, line 731: In the next sections, you added the link at the same level that you enable it but here you added in a field.
I think that this should be writted also on the `datawarehousesExportsBigQueryEnabled` field, right? :thinking:

manifests/kots-helm.yaml
```diff
@@ -104,7 +104,7 @@ spec:
       awsS3Region: '{{repl ConfigOption "storageBucketsAwsS3Region" }}'
       importAwsRoleArn: '{{repl ConfigOption "storageBucketsImportExportAwsArnRole" }}'
```
ManuelLR, line 105: This fields aren't properly configured, right? If we set the value but we hide the option in the config, the value keep placed here, right? :thinking:

manifests/kots-config.yaml
moicalcob: Not really to be honest 🤔 The next section needs more explanation because the config is more complex. That's why I added it at the top level. On this one, if the customer reads `Enable exports from BigQuery and CARTO Data Warehouse` it's easy to understand it. Once they enable it, then we show that text field with the needed link to the docs if needed, but it's not complex configuration. What do you think? 😄

manifests/kots-helm.yaml
ManuelLR: The point here is... If someone configure it in Kots config and then disable the option (so is hidden), the value keep returned and setted here, right? That's not a desired behaviour.

I'm not entirely sure if it works that way, that's why I'm asking you. I think it does since that's why we are using full conditional blocks.

moicalcob: You're totally right @ManuelLR 😄 I'm pushing the change

ManuelLR (approved):
LGTM

PR #486, opened 2024-03-04
Update Kots config to make it easier to configure imports & exports

moicalcob:


manifests/kots-config.yaml
```diff
@@ -430,57 +439,104 @@ spec:
...
+          help_text: Bucket name to be used for storing temporary files during imports
           type: text
           required: true
-          default: repl{{ if ConfigOptionEquals "replicatedLicenseType" "dev"}}repl{{ print (ConfigOption "derivedSelfHostedGCPProjectId") "-client-storage"}}repl{{ end }}
-        - name: storageBucketsWorkspaceImportsIsPublic
-          help_text: |-
-            Please select the access mode of this bucket
-          type: select_one
-          default: private
+          default: repl{{ if and (ConfigOptionEquals "replicatedLicenseType" "dev") (ConfigOptionEquals "storageBucketsBehaviour" "custom_gcs")}}repl{{ print (ConfigOption "derivedSelfHostedGCPProjectId") "-client-storage"}}repl{{ end }}
```
ManuelLR, line 461: I think that you need to add 2 conditions because in `kots-helm.yaml` you are not using the `optionalValues` block, right? :thinking:

ManuelLR (dismissed):
Looks good but I leave you a couple of comments

manifests/kots-config.yaml
```diff
@@ -23,10 +23,10 @@ spec:
         - name: backupReminder
           type: label
           title: |-
-            ⚠️  Don't forget to follow the CARTO docs regarding backups  ⚠️
-            You need to implement a backup policy for the next things in order to recover your data in case of upgrade, migration or disaster recovery:
+            ⚠️  Please, don't forget to follow CARTO docs regarding backups,
```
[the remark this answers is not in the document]
moicalcob: Done! 😄

manifests/kots-config.yaml
moicalcob: I added two because we want to put the default value just if we're in a dev license and using GCS. That's to use the default buckets that we provision with the SH 😄

ManuelLR (approved):

PR #490, opened 2024-03-08
Update config

moicalcob:


manifests/kots-config.yaml
```diff
@@ -95,8 +92,8 @@ spec:
     - name: metadataDatabase
       title: Metadata Database
       description: |-
-        Configure the information needed to connect to your metadata database.
-        If you need further info, please check [docs.carto.com](https://docs.carto.com/~/changes/1Lia4Pv0PjH33F9R0CyP/carto-self-hosted/deployment-guides/kubernetes#metadata-database).
+        Information needed to connect to your metadata database.
+        [More info](https://docs.carto.com/carto-self-hosted/key-concepts/deployment-requirements#external-database).
```
ManuelLR, line 96: ey, why you delete the custom branch from here? :thinking:

manifests/kots-config.yaml
```diff
@@ -923,21 +871,18 @@ spec:
...
       description: |-
-        Here you can config advanced options for tuning your CARTO deployment.
-        If you need further info, please check [docs.carto.com](https://docs.carto.com).
+        Advanced Configuration Section where you can enter plaintext configuration settings, provided by the CARTO support team, to tailor your installation to specific requirements.
+        
       when: repl{{ ConfigOptionEquals "enablePlatformAdvancedTuning" "1" }}
       items:
         ## Tunning
         - name: platformAdvancedTuningValues
-          help_text: |-
```
ManuelLR, line 939: I think that we should specify something like this because the client could totally break the installation putting things here and... we are not responsible of it :see_no_evil:

manifests/kots-helm.yaml
```diff
@@ -175,13 +175,13 @@ spec:
   ## Values from Advanced Configuration
   optionalValues:
     ## BigQuery exports
-    - when: '{{repl ConfigOptionEquals "datawarehousesExportsBigQueryEnabled" "1" }}'
+    - when: '{{repl not (empty (ConfigOption "storageBucketsExportGcp"))}}'
```
ManuelLR, line 178: Just so we are aware, with this approach, we will not be able to test with a **dev license** to leave the exports field undefined since it will be set to the default value so it would never be empty.

manifests/kots-config.yaml
alasarr: Just a mistake

manifests/kots-helm.yaml
moicalcob: We can just change to a trial license when required no? 😄

manifests/kots-config.yaml
```diff
@@ -923,21 +871,18 @@ spec:
...
-            ⚠️ Check only for tuning your CARTO Self-Hosted environment with advanced parameters
+
 
     # Advanced Tuning
     - name: platformAdvancedTuning
       title: Advanced Tuning
       description: |-
-        Here you can config advanced options for tuning your CARTO deployment.
-        If you need further info, please check [docs.carto.com](https://docs.carto.com).
+        Advanced Configuration Section where you can enter plaintext configuration settings, provided by the CARTO support team, to tailor your installation to specific requirements.
```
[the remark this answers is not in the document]
alasarr: I wasn't aware of this section of the doc. I think this advanced customozations that we're documenting should be in the advanced section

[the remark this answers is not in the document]
alasarr: We have agreed on removing advanced customizations from the doc.

ManuelLR (approved):
LGTM

PR #495, opened 2024-03-13
Add new preflights to check buckets and default SA

moicalcob:


chart/templates/_commonChecks.tpl
```diff
@@ -161,6 +216,57 @@ Return customer values to use in preflights and support-bundle
...
+  - name: WORKSPACE_IMPORTS_BUCKET
+    value: {{ .Values.appConfigValues.workspaceImportsBucket | quote }}
+  - name: WORKSPACE_THUMBNAILS_PROVIDER
+    value: {{ .Values.appConfigValues.storageProvider | quote }}
+  - name: WORKSPACE_IMPORTS_PROVIDER
+    value: {{ .Values.appConfigValues.storageProvider | quote }}
+  - name: WORKSPACE_THUMBNAILS_PUBLIC
+    value: {{ .Values.appConfigValues.workspaceThumbnailsPublic | quote }}
+  - name: WORKSPACE_IMPORTS_PUBLIC
+    value: {{ .Values.appConfigValues.workspaceImportsPublic | quote }}
```
ManuelLR, line 236: Why you use that value but then we have this: https://github.com/CartoDB/carto-selfhosted-helm/blob/main/manifests/kots-helm.yaml#L91 ? :thinking:

chart/templates/_commonChecks.tpl
```diff
@@ -161,6 +216,57 @@ Return customer values to use in preflights and support-bundle
...
+  - name: WORKSPACE_THUMBNAILS_BUCKET
+    value: {{ .Values.appConfigValues.workspaceThumbnailsBucket | quote }}
+  - name: WORKSPACE_IMPORTS_BUCKET
+    value: {{ .Values.appConfigValues.workspaceImportsBucket | quote }}
+  - name: WORKSPACE_THUMBNAILS_PROVIDER
+    value: {{ .Values.appConfigValues.storageProvider | quote }}
+  - name: WORKSPACE_IMPORTS_PROVIDER
+    value: {{ .Values.appConfigValues.storageProvider | quote }}
+  - name: WORKSPACE_THUMBNAILS_PUBLIC
+    value: {{ .Values.appConfigValues.workspaceThumbnailsPublic | quote }}
```
ManuelLR, line 234: Is not better to just copy this? https://github.com/CartoDB/carto-selfhosted-helm/blob/e3334154072fc058597e0ff4096250f5db650c7e/manifests/kots-helm.yaml#L93

chart/templates/_commonChecks.tpl
```diff
@@ -161,6 +216,57 @@ Return customer values to use in preflights and support-bundle
     value: {{ include "carto.postgresql.databaseName" . }}
   - name: WORKSPACE_POSTGRES_USER
     value: {{ include "carto.postgresql.user" . }}
+  - name: WORKSPACE_TENANT_ID
+    value: {{ .Values.cartoConfigValues.selfHostedTenantId | quote }}
```
ManuelLR, line 220: jummm, Today I discovered that quotes could broke things. We need to add validators to the buckets and some other places to avoid blank spaces (or properly trim all inputs)

chart/templates/_commonChecks.tpl
```diff
@@ -15,6 +15,60 @@ Return common collectors for preflights and support-bundle
             env:
             {{- include "carto.replicated.tenantRequirementsChecker.customerValues" . | indent 12 }}
             {{- include "carto.replicated.tenantRequirementsChecker.customerSecrets" . | indent 12 }}
+            volumeMounts:
+              - name: gcp-default-service-account-key
```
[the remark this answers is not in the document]
moicalcob: I followed the same approach we're using for the `workspace-api` component 🤔 When should we avoid mounting it? Just when using workload identity, no?

chart/templates/_commonChecks.tpl
moicalcob: This is at HELM chart level, not at Kots level 🙇🏼‍♂️

chart/templates/_commonChecks.tpl
moicalcob: Same as above ⬆️

chart/templates/_commonChecks.tpl
moicalcob: I think that in this case it's not dangerous 🤔 At backend level we're taking it into account so I think that there is no problem

chart/templates/_commonChecks.tpl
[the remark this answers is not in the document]
ManuelLR: Ok, put as it but... we need to dig a bit because inject 2 SA when only one is going to be used... is weird :see_no_evil:

ManuelLR (approved):

chart/templates/_commonChecks.tpl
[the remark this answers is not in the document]
moicalcob: Story created to check it: https://app.shortcut.com/cartoteam/story/395303/check-why-we-re-mounting-2-different-sas-in-containers

PR #503, opened 2024-03-20
Add alert to update doc when changing resources

moicalcob:


.github/workflows/lint-codebase.yaml
```diff
@@ -22,7 +22,46 @@ jobs:
...
+      - name: Comment on PR if Helm Chart resources changed
+        uses: peter-evans/create-or-update-comment@v2
+        if: steps.check-helm-resources-diff.outputs.resources-changed == 'true'
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            :warning: Helm Chart resources have changed. Please make sure to update the documentation accordingly. You can find more information [here](https://cartodb.getoutline.com/doc/helm-chart-yJ3vxi6icN#h-deployment-resources).
+
+      - name: Slack notification
```
ManuelLR, line 56: I think that this is unnecesary (or only required in a merge to main)

moicalcob: I've changed it to send the notification just when pushing to main 🙇🏼‍♂️ IMHO it's necessary as a comment in the PR can be omitted, but with that notification we have no excuses 😂

alvarorm22 (approved):
LGTM! 👍🏼

PR #509, opened 2024-03-26
Add preflight check for Gateway API

moicalcob:


manifests/kots-helm.yaml
```diff
@@ -19,9 +19,9 @@ spec:
       encryptionSecretKey:
         value: '{{repl ConfigOption "autogeneratedEncryptionSecretKey" }}'
       jwtApiSecret:
-        value: '{{repl ConfigOption "autogeneratedVarnishDebugSecret" }}'
-      varnishDebugSecret:
         value: '{{repl ConfigOption "autogeneratedJwtApiSecret" }}'
+      varnishDebugSecret:
```
ManuelLR, line 23: This is a breaking change so should be promoted to prod as quick as possible

moicalcob: I'm going to deploy it as soon as it's in main 😄

PR #510, opened 2024-03-27
Add application link in the Admin Console

moicalcob:


manifests/kots-app.yaml
```diff
@@ -23,3 +23,13 @@ spec:
     - deployment/carto-workspace-subscriber
     - deployment/carto-workspace-www
   minKotsVersion: 1.107.4
+---
+apiVersion: app.k8s.io/v1beta1
```
ManuelLR, line 27: Why not put in a different file? This is a standard of kubernetes not from KOTS.

Also, maybe the place is inside the Helm? :thinking:

moicalcob: It's not standard of kubernetes, and in fact this resource is not deployed as part of the chart, so we have to add it in the manifests we upload to Replicated.

I already checked this with Replicated team and they use this resource to obtain the application link, but it's not applied in the cluster as it's a CRD :)

I believe that this file is the right place as it's part of the application definition for Replicated 😄

ManuelLR (approved):

PR #512, opened 2024-04-02
Place pods in different nodes when enabling high availability

moicalcob:


ManuelLR (approved):
Let's go !

After that... should we put a warning regarding number of nodes when high availability is enabled? :thinking:

moicalcob:
@ManuelLR I've checked that with just one node it works correctly in the embedded cluster, so I think that it's not really needed, no? 😄

PR #520, opened 2024-04-11
Provide encryption keys from CARTO side by default

moicalcob:


manifests/kots-config.yaml
```diff
@@ -903,6 +891,27 @@ spec:
           required: true
           default: "3"
 
+        ## Custom Encryption Keys
+        - name: enableCustomEncryptionKeys
```
ManuelLR, line 895: But we said that this option is going to be hidden and only enabled to some clients by demand, right? :thinking:

moicalcob: Oh yes 🙇🏼‍♂️ Fixed!

ManuelLR (approved):

PR #521, opened 2024-04-12
Add security context and limit resources in Preflights component

alvarorm22:
**Description of the change**

- Add security context to pod and containers in preflights component

Platform distribution tested:
- [x] GKE
- [x] EKS
- [x] OpenShift  

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**


chart/values.yaml
```diff
@@ -4587,6 +4587,16 @@ tenantRequirementsChecker:
...
+  ## Configure Pods Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
+  ## @param tenantRequirementsChecker.podSecurityContext.enabled Enabled Upgrade Check pre-hook pods' Security Context
+  ## @param tenantRequirementsChecker.podSecurityContext.fsGroup Set Upgrade Check pre-hook pod's Security Context fsGroup
+  ## @param tenantRequirementsChecker.podSecurityContext.supplementalGroups[0] Set Upgrade Check pre-hook pod's Security Context supplementalGroups
+  ##
+  podSecurityContext:
+    enabled: true
+    fsGroup: 101
+    supplementalGroups: [2345]
```
moicalcob, line 4599: I think that we're not using this in the pod template no? 🤔 Same happens with the next and previous lines? 😓

alvarorm22: Yep, you are right, I think the best approach is defining these in the values and then referring them in the preflights, I will do these changes

moicalcob (approved):
LGTM! Thanks!

PR #528, opened 2024-04-29
fix proxy connection string

alvarorm22:
**Description of the change**

- Proxy connection string should be with `<protocol>://<host>:<port>` by default
- We should be able to override this value with the parameter `externalProxy.connectionString`
**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

### DIFF
--- a/manifests/kots-helm.yaml
+++ b/manifests/kots-helm.yaml
@@ -289,7 +289,8 @@ spec:
       values:
         externalProxy:
           enabled: true
-          connectionString: '{{repl ConfigOption "externalProxyHost" }}'
+          type: '{{repl ConfigOption "externalProxyType" }}'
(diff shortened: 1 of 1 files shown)


manifests/kots-helm.yaml
```diff
@@ -289,7 +289,8 @@ spec:
       values:
         externalProxy:
           enabled: true
-          connectionString: '{{repl ConfigOption "externalProxyHost" }}'
+          type: '{{repl ConfigOption "externalProxyType" }}'
+          host: '{{repl ConfigOption "externalProxyHost" }}'
```
moicalcob, line 293: Why do we need this? It should work with the `connectionString` property, no? 🤔

PR #544, opened 2024-06-03
breaking change: update .Values.tlsCerts and ingress resources

alvarorm22:
**Description of the change**

- [x] Update TLS Certificates configuration
- [x] Update Ingress resources 

## TLS Certificates

With these changes, we are going from
```
tlsCerts:
  httpsEnabled: false
  autoGenerate: false
  existingSecret:
    name: ""
    keyKey: ""
    certKey: ""
```
to:
```
router:
  httpsEnabled: false
  tlsCertificates:
    certificateValueBase64: ""
    privateKeyValueBase64: ""
    existingSecret:
      name: ""
      certKey: ""
      keyKey: ""
```

```
gateway:
  tlsCertificates:
    customSSLCerts:
      enabled: false
      certificateValueBase64: ""
      privateKeyValueBase64: ""
    managedCerts:
      enabled: false
      name: ""
```

```
ingress:
  tls: false
  tlsCertificates:
    existingSecret:
      name: ""
      certKey: ""
      keyKey: ""
```

Each component has their own TLS and certificate configurations and the `.Values.tlsCerts` key will be deleted

## Ingress
From now on, the **ingress** resources are not inside the router component, they are below the key `.Values.ingress`, including the TLS option and certificate secret reference

If a customer that has now the Carto Platform deployed with ingress wants to use it also with Replicated, they must follow these steps:

- Select **_Custom Access Mode_**
- Select "**_Terminate TLS in a higher layer and connect to Carto over HTTP_**"
- Add this config inside Custom Configuration:

```
ingress:
  enabled: true
  tls: true
  existingSecret:
    name: <domain tls secret reference> # It must exist in the namespace
router:
  service:
    type: NodePort
    annotations:
      cloud.google.com/neg: '{"ingress": true}'
extraDeploy:
  - |
    ## FrontendConfig object is used to configure downstream Ingress features
    ## Redirect port 80 to 443 automatically
    ## https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#https_redirect
    apiVersion: networking.gke.io/v1beta1
    kind: FrontendConfig
    metadata:
      name: carto-ingress-frontend-config
      labels: {{- include "common.labels.standard" . | nindent 4 }}
        app.kubernetes.io/component: carto
        {{- if .Values.commonLabels }}
        {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
        {{- end }}
      annotations:
        {{- if .Values.commonAnnotations }}
        {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
        {{- end }}
      namespace: {{ .Release.Namespace | quote }}
    spec:
      redirectToHttps:
        enabled: true
      ## Attach a SSL Policy
      # sslPolicy: my-ssl-policy
```


alvarorm22:
https://app.shortcut.com/cartoteam/story/415183/check-ingress-resources-as-part-of-breaking-change

alvarorm22:
https://app.shortcut.com/cartoteam/story/388886/check-tlscerts-httpsenabled-and-tlscerts-autogenerate-helm-chart-options

chart/templates/router/ingress.yaml
moicalcob, line 1: Shouldn't the ingress be located in a different folder now that it's not attached to the router? 🤔

chart/templates/NOTES.txt
moicalcob, line 1: Why are we deleting this on? 😮 From now on we won't show that to customers?

chart/templates/router/ingress.yaml
alvarorm22: yep, makes sense

chart/templates/NOTES.txt
alvarorm22: do you think it is useful for customers? if yes I can update and adapt it because it is obsolete

alvarorm22: updated adding the gateway option and link to the new doc

PR #545, opened 2024-06-04
Add CI to deploy changes generated in a branch

moicalcob:


.github/workflows/branch-changes-release.yaml
```diff
@@ -0,0 +1,63 @@
...
+          export NORMALIZED_GIT_BRANCH_NAME=${GITHUB_BRANCH_NAME//[^[:alnum:]]/_}
+
+          # To lowercase all characters
+          export NORMALIZED_GIT_BRANCH_NAME=${NORMALIZED_GIT_BRANCH_NAME,,}
+
+          # Set maximum 128 characters for image name
+          export NORMALIZED_GIT_BRANCH_NAME=${NORMALIZED_GIT_BRANCH_NAME:0:127}
+
+          # Share the variable for further steps
+          echo "channel-name=${NORMALIZED_GIT_BRANCH_NAME}" >> "$GITHUB_OUTPUT"
```
alvarorm22, line 40: do you know if replicated has a max length for the channel name?

moicalcob: No 🙇🏼 I added that limit as it's the same we're using in cloud-native for the docker images and looked okay for me just in case 😄

PR #549, opened 2024-06-05
Add readOnlyRootFilesystem to helm chart

moicalcob:


alvarorm22:
https://app.shortcut.com/cartoteam/story/415856/router-component-should-use-an-emptydir-volume-to-generate-the-templates-used-by-nginx

chart/templates/accounts-www/deployment.yaml
```diff
@@ -144,13 +144,21 @@ spec:
...
+              mountPath: /tmp
           {{- if .Values.accountsWww.extraVolumeMounts }}
           {{- include "common.tplvalues.render" (dict "value" .Values.accountsWww.extraVolumeMounts "context" $) | nindent 12 }}
           {{- end }}
         {{- if .Values.accountsWww.sidecars }}
         {{- include "common.tplvalues.render" ( dict "value" .Values.accountsWww.sidecars "context" $) | nindent 8 }}
         {{- end }}
       volumes:
+        - name: config
+          emptyDir: {}
```
ManuelLR, line 159: After checking https://kubernetes.io/docs/concepts/storage/volumes/#emptydir looks like the medium used is the default configured by each cluster, not always memory.

What do you think if we explicitly define as "memory" for the critical ones? :thinking: 
I'm not 100% sure if that could cause some problems in client side in case they fully disable the option to use "memory" volumes (and also I don't know if could be done). Discuss with Alvaro if you have doubts :pray: .

alvarorm22: Ok go ahead with memory for config volumes

alvarorm22 (dismissed):
LGTM 👍🏼

ManuelLR (approved):
LGTM

ManuelLR (approved):

PR #563, opened 2024-06-14
Add ingress testing mode for Kots installation

moicalcob:


chart/templates/_helpers.tpl
```diff
@@ -1354,43 +1354,6 @@ Return YAML for the Redis init container
...
-{{- $messages := append $messages (include "carto.validateValues.redis" .) -}}
-{{- $messages := append $messages (include "carto.validateValues.postgresql" .) -}}
-{{- $messages := without $messages "" -}}
-{{- $message := join "\n" $messages -}}
-
-{{- if $message -}}
-{{-   printf "\nVALUES VALIDATION:\n%s" $message | fail -}}
-{{- end -}}
-{{- end -}}
-
```
moicalcob, line 1393: I've moved this to the validators template that we have 😄

chart/templates/_helpers.tpl
```diff
@@ -1354,43 +1354,6 @@ Return YAML for the Redis init container
...
-If internalRedis.enabled=false you need to specify the host of an external Redis(TM) instance setting externalRedis.host
-{{- end -}}
-{{- end -}}
-
-{{/*
-Validate external Redis config
-*/}}
-{{- define "carto.validateValues.postgresql" -}}
-{{- if and (not .Values.internalRedis.enabled) (not .Values.externalRedis.host) -}}
```
moicalcob, line 1372: Changes in the new location. We were using redis values to validate postgres 😓

chart/templates/_helpers.tpl
```diff
@@ -1354,43 +1354,6 @@ Return YAML for the Redis init container
...
-{{- define "carto.validateValues.redis" -}}
-{{- if and (not .Values.internalRedis.enabled) (not .Values.externalRedis.host) -}}
-CARTO: Missing Redis(TM)
-
-If internalRedis.enabled=false you need to specify the host of an external Redis(TM) instance setting externalRedis.host
-{{- end -}}
-{{- end -}}
-
-{{/*
-Validate external Redis config
```
moicalcob, line 1369: Changes, we were not validating Redis here but Postgres

chart/values.yaml
```diff
@@ -75,14 +75,20 @@ appConfigValues:
...
   ldsIsolineProvider: ""
 
-## @section CARTO config parameters
-## Global configuration provided by CARTO. If you changed something from this section probably your self-hosted is going to stop working.
+## @section CARTO Replicated parameters
+## Global configuration for installations using Replicated. If you change something from this section probably your self-hosted is going to stop working.
 replicated:
   ## @param replicated.enabled Enable Carto replicated
   ## @param replicated.platformDistribution Platform distribution to deploy Carto Selfhosted with Replicated
+  ## @param replicated.ingressTestingMode Enable router testing mode for the application. This will just deploy the router in testing mode serving a simple page to check that it works.
```
ManuelLR, line 83:
```suggestion
  ## @param replicated.ingressTestingMode Enable router testing mode for the application. This will deploy the router in testing mode serving a simple page to check that it works without routing traffic to other components..
```

chart/values.yaml
```diff
@@ -75,14 +75,20 @@ appConfigValues:
...
+  ## @param replicated.onlyRunRouter Enable only the router component in the installation. This will just deploy the router component to check the ingress layer.
```
ManuelLR, line 84:
```suggestion
  ## @param replicated.onlyRunRouter Enable only the router component in the installation. This will just deploy the router component. Useful to check the ingress layer together with the option `replicated.ingressTestingMode`.
```

manifests/kots-config.yaml
```diff
@@ -442,6 +442,15 @@ spec:
             If you don't provide an Allocation ID, a random domain will be assigned to your load balancer. You can find some information about how to use the domain you configured for CARTO platform in the [AWS ELB documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/using-domain-names-with-elb.html).
           recommended: true
 
+        # Ingress testing mode
+        - name: enableIngressTestingMode
+          title: Ingress testing mode
+          when: '{{repl (ConfigOptionEquals "accessToCartoModeK8s" "accessToCartoModeK8sCustom") }}'
```
ManuelLR, line 448: I thought we were going to leave it available for all modes.

I don't have a strong opinion but you should also think about the single instance that places a load balancer in front of the instance :see_no_evil:

moicalcob: It's configured right now to be shown in both embedded cluster and existing cluster 🤔 Maybe the `accessToCartoModeK8s` is confusing as it handles both the embedded cluster and the existing cluster

moicalcob: 😓 You were right, I missed one part of the conditional 😅 Fixed! Thanks for checking it!

alvarorm22 (dismissed):
That's Ok for me and can be very useful 👌🏼

RuBiCK (dismissed):
LGTM

manifests/kots-config.yaml
```diff
@@ -432,6 +432,15 @@ spec:
             If you don't provide an Allocation ID, a random domain will be assigned to your load balancer. You can find some information about how to use the domain you configured for CARTO platform in the [AWS ELB documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/using-domain-names-with-elb.html).
           recommended: true
 
+        # Ingress testing mode
+        - name: enableIngressTestingMode
+          title: Ingress testing mode
+          when: '{{repl or (ConfigOptionEquals "accessToCartoModeK8s" "accessToCartoModeK8sCustom") (ConfigOptionEquals "accessToCartoModeK8sNotFullySupported" "accessToCartoModeK8sCustom") (ConfigOptionEquals "isEmbeddedCluster" "true") }}'
+          help_text: |-
+            Enable if you want to enable ingress testing mode. This will just deploy the CARTO router in testing mode to check that the layers you place on top of CARTO work.
```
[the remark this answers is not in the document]
moicalcob: Exactly 😄 It will show a dummy website. The changes are added in the router-http image here: https://github.com/CartoDB/cloud-native/pull/16898 Basically we change the Nginx config to route all requests to a static website.

The layers that can be placed on top of CARTO are on the customer side. We offer the default ingress method were you don't have to worry about anything and the custom method.

In the custom method, customer can terminate the TLS at the router level providing their certificate and key, or they can serve traffic using HTTP and terminate the TLS at a higher layer.

In that particular scenario is where this ingress testing mode helps, as we'll be able to ask customer to do a port-forward to the router service and if the static website appear, then they have to check whatever they're putting on top of the SH if it doesn't work :)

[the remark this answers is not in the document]
RuBiCK: A suggestion, or something similar
```suggestion
            Enable if you want to turn on ingress testing mode. This will deploy the CARTO router in testing mode, which shows a hello-world page, without redirecting traffic to the internal components, to check that the layers you place on top of CARTO work.
```

PR #564, opened 2024-06-17
update replicated sdk to be able to use proxy

alvarorm22:
**Description of the change**

- [x] Add proxy env vars to Replicated SDK when proxy is enabled

### DIFF
--- a/manifests/kots-helm.yaml
+++ b/manifests/kots-helm.yaml
@@ -310,6 +310,12 @@ spec:
           sslRejectUnauthorized: repl{{ if ConfigOptionEquals "externalHttpsProxySslCertificateCheck" "0"}}falserepl{{ else }}truerepl{{ end }}
           sslCA: |
             repl{{ ConfigOptionData "externalHttpsProxySslCa" | nindent 12 }}
+        replicated:
+          extraEnv:
+          - name: repl{{ if ConfigOptionEquals "externalProxyType" "http"}}HTTP_PROXYrepl{{ else }}HTTPS_PROXYrepl{{ end }}
+            value: '{{repl ConfigOption "externalProxyHost" }}'
+          - name: NO_PROXY
+            value: '{{repl ConfigOption "externalProxyExcludedDomains" }}'
 
     - when: '{{repl ConfigOptionEquals "ssoEnabled" "1"}}'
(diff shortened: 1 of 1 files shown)


alvarorm22:
https://app.shortcut.com/cartoteam/story/412706/check-replicated-sdk-behaviour-with-external-proxy-configured

manifests/kots-helm.yaml
```diff
@@ -310,6 +310,12 @@ spec:
           sslRejectUnauthorized: repl{{ if ConfigOptionEquals "externalHttpsProxySslCertificateCheck" "0"}}falserepl{{ else }}truerepl{{ end }}
           sslCA: |
             repl{{ ConfigOptionData "externalHttpsProxySslCa" | nindent 12 }}
+        replicated:
+          extraEnv:
+          - name: repl{{ if ConfigOptionEquals "externalProxyType" "http"}}HTTP_PROXYrepl{{ else }}HTTPS_PROXYrepl{{ end }}
+            value: '{{repl ConfigOption "externalProxyHost" }}'
```
moicalcob, line 316: We have to add the value in both `HTTP_PROXY` env var and `HTTPS_PROXY` env var just in case. That variable doesn't mean that the proxy is http/https, but that for http and https requests the proxy will be used using the protocol specified by the proxy. So if I'm not wrong we have to add it for both protocols

moicalcob (changes requested):

manifests/kots-helm.yaml
alvarorm22: I've tested again with the two vars and the `replicated-sdk` pod throws the same error that last day:
```
2024/06/20 14:39:45 failed to bootstrap, retrying in 10s: failed to get latest license: failed to get license from api: failed to execute get request: Get "https://replicated.self-hosted.carto.com/license/carto": malformed HTTP response "\x15\x03\x03\x00\x02\x01\x00"
```

It has only worked for me by adding one of the two variables

moicalcob: 🤔 Maybe replicated expects just one configured in all pieces? That sounds weird to me as it's not the "standard" approach so we could maybe open a support request to check it?

moicalcob (approved):

PR #566, opened 2024-06-21
enable supportMinimalRBACPrivileges and cluster preflights

alvarorm22:
**Description of the change**

- [x] Allow using namespace RBAC instead of cluster role
- [x] Update cluster preflights to run only with the cluster role
- [x] Update Access Mode in Kots-config panel

### DIFF
--- a/chart/templates/_commonChecks.tpl
+++ b/chart/templates/_commonChecks.tpl
@@ -233,8 +233,7 @@ Return common analyzers for preflights and support-bundle
             message: Failed to check if images are present in registry
         - pass:
             message: All Carto images are available
-  "{{ printf "{{- $role := lookup rbac.authorization.k8s.io/v1 Role .Release.Namespace kotsadm-role }}" }}"
-  "{{ printf "{{- if not $role }}" }}"
+  {{- if not (lookup "rbac.authorization.k8s.io/v1" "Role" .Release.Namespace "kotsadm-role") }}
   - clusterVersion:
       outcomes:
         - fail:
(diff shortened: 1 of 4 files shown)


alvarorm22:
https://app.shortcut.com/cartoteam/story/420736/allow-using-namespace-rbac

chart/templates/_commonChecks.tpl
```diff
@@ -233,6 +233,7 @@ Return common analyzers for preflights and support-bundle
             message: Failed to check if images are present in registry
         - pass:
             message: All Carto images are available
+  {{- if ne .Values.replicated.platformDistribution "" }}
```
moicalcob, line 236: Can we please add a comment here explaining why we have this? 🙏🏼

manifests/kots-config.yaml
```diff
@@ -349,10 +349,12 @@ spec:
               title: |-
                 Custom
         - name: accessToCartoModeK8sNotFullySupported
+          hidden: true
```
moicalcob, line 352: Why is this one hidden? 🤔 We are rendering it just when needed no?

moicalcob: Oh got it, makes sense to me! Thanks!

alvarorm22: Yes, but I think that it is a bit strange to show only the "Custom" option if it is the only possibility to select 🤔

chart/templates/_commonChecks.tpl
```diff
@@ -233,6 +233,11 @@ Return common analyzers for preflights and support-bundle
             message: Failed to check if images are present in registry
         - pass:
             message: All Carto images are available
+  {{/*
+  We only can run the following preflight checks and get the platform distribution when a cluster role is created.
+  Otherwise, we can't obtain this info
```
moicalcob, line 238:
```suggestion
  Otherwise, we cannot obtain this info
```
It's probably going to cause some issues having that single quote there

PR #615, opened 2024-09-19
allow to use proxy https with replicated

alvarorm22:
**Description of the change**

- [x] Be able to use Proxy HTTPS with a Private TLS Cert


moicalcob (dismissed):
Nice job!!! 😄 🚀

PR #616, opened 2024-09-19
delete old proxy env vars in replicated sdk

alvarorm22:
**Description of the change**

The values provided to the `--http-proxy`, `--https-proxy`, and `--no-proxy` flags for the kots install command are now propagated to the Replicated SDK, so there is no need to add them manually.


Keralin (approved):
LGTM

moicalcob (approved):

PR #643, opened 2024-12-08
Replace router-metrics for a better alternative consuming less resources

moicalcob:


RuBiCK (dismissed):

chart/values.yaml
```diff
@@ -5046,10 +5040,10 @@ routerMetrics:
   ##
   resources:
     limits:
-      memory: "1464Mi"
+      memory: "1024Mi"
```
ManuelLR, line 5043:
```suggestion
      memory: "512Mi"
```

ManuelLR (dismissed):

PR #650, opened 2024-12-17
Include support bundle & preflight checks in helm installation

moicalcob:
Create the support bundle and the preflights secret in both the Kots and the Helm installation by default, and adapted the checks to use existing secrets in the Helm chart deployment. Now we support existing secrets for:

- `cartoSecrets.defaultGoogleServiceAccount`
- `appSecrets.awsAccessKeyId`
- `appSecrets.awsAccessKeySecret`
- `appSecrets.azureStorageAccessKey`
- `appSecrets.googleCloudStorageServiceAccountKey`
- `externalPostgresql`
- `externalRedis`
- `launchDarklySdkKey`

### DIFF
--- a/chart/templates/_commonChecks.tpl
+++ b/chart/templates/_commonChecks.tpl
@@ -62,13 +62,29 @@ Return common collectors for preflights and support-bundle
                 done
             env:
               {{- if not .Values.commonBackendServiceAccount.enableGCPWorkloadIdentity }}
+              {{- if eq .Values.cartoSecrets.defaultGoogleServiceAccount.existingSecret.name "" -}}
               - name: DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT
                 value: {{ .Values.cartoSecrets.defaultGoogleServiceAccount.value | b64enc | quote }}
+              {{- else -}}
+              - name: DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT
+                valueFrom:
+                  secretKeyRef:
+                    name: {{ .Values.cartoSecrets.defaultGoogleServiceAccount.existingSecret.name | quote }}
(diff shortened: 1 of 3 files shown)


chart/templates/_commonChecks.tpl
```diff
@@ -510,28 +526,72 @@ Return customer values to use in preflights and support-bundle
...
+  {{- end -}}
+  {{- if eq .Values.cartoSecrets.launchDarklySdkKey.existingSecret.name "" }}
   - name: LAUNCHDARKLY_SDK_KEY
     value: {{ .Values.cartoSecrets.launchDarklySdkKey.value | quote }}
-  {{- if eq .Values.appConfigValues.storageProvider "s3" }}
+  {{- else -}}
+  {{ include "carto._utils.generateSecretDef" (dict "var" "LAUNCHDARKLY_SDK_KEY" "context" .) | nindent 2 }}
+  {{- end -}}
+  {{- if eq .Values.appConfigValues.storageProvider "s3" -}}
+  {{- if eq .Values.appSecrets.awsAccessKeyId.existingSecret.name "" }}
```
ManuelLR, line 556: A trick for the future to improbe readability, you can add additional whitespaces to properly read it. Example:
```
{{- if eq .Values.appConfigValues.storageProvider "s3" -}}
{{-   if eq .Values.appSecrets.awsAccessKeyId.existingSecret.name "" }}
      - xyz
{{-   end -}}
{{-   if eq .Values.appSecrets.awsAccessKeyId.existingSecret.name "" }}
      - xyz
{{-   end -}}
{{- end -}}
```

ManuelLR (approved):

chart/templates/_commonChecks.tpl
moicalcob: If I add these additional whitespaces then the template is generated with them 😓 I'm not sure if there is any way we can do it 🤔

PR #664, opened 2025-02-10
chore(replicated sdk): Update lockfile for the helm chart

moicalcob:


mateo-di (approved):
Checked with Moises vía Slack. Result of the 'helm dependency update'  command execution

PR #690, opened 2025-03-28
Update proxy host validation regex in kots-config.yaml

moicalcob:


miguelangelmorenochacon (approved):
No idea what that regex does but go

PR #697, opened 2025-04-15
warn using a list and add the conditionals. 1st iteration

mateo-di:
**Description of the change**

Marks some preflight checks as warnings instead of failures when related features (e.g., TravelTime, TomTom, buckets) are not configured in `values.yaml`. Uses a list to track optional checks.

**Benefits**

- Avoids false negatives in minimal setups.
- Clearer UX for users not using all external services.
- Simplifies logic with a cleaner list-based approach.

**Possible drawbacks**

- Only supports a single optional level (`warn`), not more granular outcomes.

**Applicable issues**

- fixes #1827

**Additional information**

Tested with and without optional values to confirm expected `warn` vs `fail` outcomes.

### DIFF
--- a/chart/templates/_commonChecks.tpl
+++ b/chart/templates/_commonChecks.tpl
@@ -261,6 +261,26 @@ NOTE: Remember that with the ingress testing mode the components are not deploye
       "EgressRequirementsValidator" (list "Check_CARTO_Auth_connectivity" "Check_PubSub_connectivity" "Check_Google_Storage_connectivity" "Check_release_channels_connectivity" "Check_Google_Storage_connectivity" "Check_CARTO_images_registry_connectivity" "Check_TomTom_connectivity" "Check_TravelTime_connectivity")
       "PubSubValidator" (list "Check_publish_and_listen_to_topic")
   }}
+{{/* Add optional analyzers to the preflightsDict */}}
+
+{{- $preflightOptionalList := list }}
+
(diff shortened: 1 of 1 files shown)


moicalcob (changes requested):

chart/templates/_commonChecks.tpl
```diff
@@ -261,6 +261,16 @@ NOTE: Remember that with the ingress testing mode the components are not deploye
...
       "PubSubValidator" (list "Check_publish_and_listen_to_topic")
   }}
+  
+  {{/* Add optional analyzers to the preflightsDict */}}
+
+  {{- $preflightOptionalList := list
+      "Check_TravelTime_connectivity"
+      "Check_TomTom_connectivity"
+      "Check_temp_bucket"
+      "Check_assets_bucket"
```
moicalcob, line 271: Hay @mateo-di these two are mandatories :)

moicalcob (changes requested):

moicalcob (approved):
LGTM! 🚀 Thanks so much 😄 It would be great if you could add a screenshot of the result in the PR description for future PRs editing the preflight checks 🙏🏼

mateo-di:
> LGTM! 🚀 Thanks so much 😄 It would be great if you could add a screenshot of the result in the PR description for future PRs editing the preflight checks 🙏🏼

Done Moises! Thanks for the help :) 
image

PR #737, opened 2025-07-14
Fully support external HTTP and HTTPs proxies with KOTS

moicalcob:


chart/templates/_helpers.tpl
```diff
@@ -1286,7 +1290,11 @@ Return the directory where the proxy CA cert will be mounted
 Return the filename where the proxy CA will be mounted
 */}}
 {{- define "carto.proxy.configMapMountFilename" -}}
+{{- if .Values.externalProxy.sslCA -}}
 {{- print "ca.crt" -}}
+{{- else if .Values.externalProxy.sslCAConfigmapName -}}
+{{- print "ca_0.crt" -}}
```
moicalcob, line 1296: Confirming with Replicated team is that's the right approach here https://cartoteam.slack.com/archives/C05V5V84W7M/p1752499969534919

chart/values.yaml
```diff
@@ -5141,6 +5141,8 @@ externalProxy:
   sslRejectUnauthorized: true
   ## @param externalProxy.sslCA CA for the proxy SSL certificate in case is self-signed or signed by a not well-known CA
   sslCA: ""
+  ## @param externalProxy.sslCAConfigmapName Name of the ConfigMap containing the CA certificate for the proxy SSL certificate
+  sslCAConfigmapName: ""
```
ManuelLR, line 5145: You should also define the key of the configmap, right?

Also, is that our standard way when we allow pass the value/secret or reference it from an external place? (like pass the raw secrets value vs secretName+key)

Please, keep thinking first on helm and then on Kots, in that way we can create Helm package usable without breaking replicated things.

manifests/kots-app.yaml
```diff
@@ -24,9 +24,9 @@ spec:
     - deployment/carto-workspace-subscriber
     - deployment/carto-workspace-www
     - '{{repl if ConfigOptionEquals "externalRedisEnabled" "0" }}deployment/carto-redis{{repl else}}{{repl end }}'
-  minKotsVersion: 1.107.4
+  minKotsVersion: 1.117.3
   additionalImages:
-    - gcr.io/carto-onprem-artifacts/tenant-requirements-checker:2024.4.17-rc.7
+    - gcr.io/carto-onprem-artifacts/tenant-requirements-checker:2025.6.10
```
ManuelLR, line 29: Do we have automatized in some way the auto-update of that image version on that file? if not it could cause us problems 🙈

chart/templates/_commonChecks.tpl
```diff
@@ -218,6 +218,11 @@ Return common collectors for preflights and support-bundle
             emptyDir:
               sizeLimit: 8Mi
           {{- end }}
```
ManuelLR, line 220: I suppose that here you should add a check to avoid define the secret as value and as configmap as the same time, right?

chart/values.yaml
moicalcob: Ah, much better! Yes it totally makes sense, I'm checking with replicated team how to handle the key that is used in the secret. I'll update it ASAP, thanks!

My initial idea was to force customers to create a secret identical to the one that Replicated creates to avoid issues but it's much easier if we allow defining the key and we hardcode it when using KOTS

chart/templates/_commonChecks.tpl
moicalcob: It's already checked :) If we define `.Values.externalProxy.sslCA` we define it as value, but if `.Values.externalProxy. sslCAConfigmapName` is defined then we define it as configmap

moicalcob: Just added it, nice cath :)

ManuelLR (approved):

PR #770, opened 2025-09-19
Hide aiFeatures until GA

mateo-di:


mateo-di:
PD: not sure why AiApi was failing with no EVENT_BUS_TOPIC EnvVar. Need to investigate this, this is not happening anywere else. Only happened to me for this testing.

moicalcob:
> PD: not sure why AiApi was failing with no EVENT_BUS_TOPIC EnvVar. Need to investigate this, this is not happening anywere else. Only happened to me for this testing.

@mateo-di let's please create a story to check it 🙏🏼 We probably need a new env var now that they're sending events from the AI API

moicalcob (approved):

PR #849, opened 2026-04-23
feat(chart): add aiApi.customHeaders for LLM gateway headers [sc-544037]

mateo-di:
**Description of the change**

Adds a dedicated `aiApi.customHeaders` value that the chart renders as the `AI_CUSTOM_HEADERS` JSON env var into the ai-api configmap. The env var is only emitted when the map is non-empty, so the default behavior is unchanged.

ai-api support for this env var shipped in CartoDB/cloud-native#23965 but nothing in the helm chart was injecting it — self-hosted deployments fronted by a corporate LLM gateway need it to pass authentication/tracking headers through LiteLLM to their provider.

Story: [sc-544037](https://app.shortcut.com/cartoteam/story/544037)

**Benefits**

- First-class, discoverable parameter in `values.yaml` / README instead of forcing operators to craft an `aiApi.extraEnvVars` entry with a hand-escaped JSON string.
- Operators write a normal YAML map; the chart handles `toJson`.
- Template variables (`{{sso_user_id}}`, `{{client_ip}}`, `{{user_agent}}`, `{{account_id}}`, `{{ai_feature}}`, `{{ai_proxy_ip}}`) are resolved per-request by ai-api.

**Possible drawbacks**

None — the env var is only rendered when `aiApi.customHeaders` is non-empty, and ai-api itself treats it as optional (self-hosted only, falls back to the per-account DB column when unset).

**Applicable issues**

- Pairs with CartoDB/cloud-native#23965 (ai-api side — already merged).

**Additional information**

**Validation**

`helm template` with `aiApi.customHeaders.X-Test=hello` renders the configmap with:
```
AI_CUSTOM_HEADERS: "{\"X-Test\":\"hello\"}"
```
Omitting the value renders a configmap without `AI_CUSTOM_HEADERS`, matching current behavior.

README was regenerated by the `helm-readme-generator` pre-commit hook (column widths realigned as a side effect of the new row).


josemaria-vilaplana (approved):
LGTM

PR #854, opened 2026-04-26
fix(chart): redact sensitive values from preflight + support-bundle artifacts [sc-569079]

mateo-di:
## Summary

CARTO support bundles were leaking customer secrets in plaintext. This PR adds a redactor block to the preflight and support-bundle specs that scrubs every known leak shape before the `tar.gz` is written.

Replicated Troubleshoot applies `spec.redactors:` in-stream during the `collect → redact → archive` pipeline, *before* the bundle is written to disk and before anything leaves the cluster. So inline values inside the rendered podSpec / Secret are fine as long as the redactor set covers their shape in the captured artifact.

This is the **primary leak control**, layered on top of Replicated's built-in redactors that already handle `*PASSWORD*`, `*TOKEN*`, `*ACCESS_KEY_ID`, `*SECRET_ACCESS_KEY`, `*OWNER_ACCOUNT`, and `protocol://user:pass@host` connection strings.

## Why this approach (and why we walked back the earlier `secretKeyRef` refactor)

The earlier commits in this branch swapped inline `value:` to `secretKeyRef` so the snapshotted PodSpec in support bundles would only show Secret names. That works in pure-helm install — but **breaks fresh installs in the Replicated / KOTS flow**: KOTS runs preflights *pre-install*, before any chart-generated Secret exists in the cluster. The runPod would fail with `CreateContainerConfigError`, sit Pending for the 180s timeout, and KOTS would block the deploy.

This PR restores the original inline-value paths (zero `valueFrom` refs in the preflight runPod, KOTS pre-install works) and concentrates the leak control entirely in the redactors block.

## Diff scope

```
chart/templates/_commonChecks.tpl   | +142  (new define carto.replicated.commonChecks.redactors at end of file)
chart/templates/preflight.yaml      |   +2  (include redactors)
chart/templates/support-bundle.yaml |   +4  (include redactors)
```

The rest of `_commonChecks.tpl` and `_helpers.tpl` are byte-identical to main. No `customerSecrets` refactor, no `generateSecretDef` extension, no behavior changes to the existing collectors / analyzers.

## Redactor set — designed against real bundles

Audited **24 real customer support bundles** on disk to derive the leak surface. Findings drove the pattern set.

### Leak surfaces observed

| File | Field / env var | Real shape | Coverage |
|---|---|---|---|
| `tenant-requirements-check/tenant-requirements-check.json` | `DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT` | base64 of GCP SA JSON with `private_key` (~3000+ chars) | `gcp-sa-json-base64` |
| same | `STORAGE_SERVICE_ACCOUNT_KEY__FILE_CONTENT` | same | `gcp-sa-json-base64` |
| same | `ROUTER_SSL_CERT_KEY__FILE_CONTENT` | base64 of `-----BEGIN RSA PRIVATE KEY-----` | `pem-private-keys-base64` |
| same | `WORKSPACE_*_STORAGE_ACCESSKEY` | Azure 88-char base64 (no shape) | `tenant-requirements-check-env-fallback` (`yamlPath`) |
| same | `WORKSPACE_POSTGRES_PASSWORD`, `*_ACCESSKEYID`, `*_SECRETACCESSKEY` | — | already redacted by KOTS built-ins |
| `replicated-sdk/*/replicated-license-info-stdout.txt` | `cartoPlatformDefaultSA` | full GCP SA JSON with `-----BEGIN PRIVATE KEY-----` | `pem-private-keys-raw` + `gcp-sa-private-key-field` |
| same | `openAiApiKey` | `sk-svcacct-...` | `openai-sk-keys` |
| same | `geminiApiKey` | `AIzaSy...` | `google-api-keys` |
| same | `vitallyToken` | base64-wrapped JWT (`ZXlK...`) | `jwt-tokens-base64-wrapped` |
| same | `cartoFeaturesFlagSdkKey` (LaunchDarkly) | `sdk-uuid` | `launchdarkly-sdk-keys` + `carto-license-sensitive-entitlements` |
| same | `databaseEncryptionKey`, `jwtEncryptionKey`, `instanceId`, `cartoAuthClientId` | shape-less | `carto-license-sensitive-entitlements` (scoped to license-info files only) |

### 13 redactor rules

| Name | Catches |
|---|---|
| `gcp-sa-json-base64` | base64-wrapped GCP service account JSON object |
| `gcp-sa-private-key-field` | `"private_key": "-----BEGIN..."` inside any JSON |
| `pem-private-keys-raw` | `-----BEGIN ... PRIVATE KEY-----` blocks |
| `pem-private-keys-base64` | base64-encoded PEM private keys (`LS0tLS1CRUdJTi...`) |
| `launchdarkly-sdk-keys` | `sdk-<uuid>` |
| `launchdarkly-mobile-keys` | `mob-<uuid>` |
| `aws-access-key-ids` | `AKIA…` and `ASIA…` |
| `openai-sk-keys` | `sk-`, `sk-svcacct-`, `sk-proj-` |
| `google-api-keys` | `AIzaSy...` |
| `jwt-tokens` | `header.payload.sig` |
| `jwt-tokens-base64-wrapped` | base64 of a JWT (`ZXlK...`) |
| `carto-license-sensitive-entitlements` | catch-all over named entitlement fields, **scoped to the license-info file** so it doesn't redact unrelated `"value":"..."` JSON elsewhere |
| `tenant-requirements-check-env-fallback` | `yamlPath` over every env value in the tenant-requirements-check pod, **scoped via fileSelector** so debug bundles keep `LOG_LEVEL`, `NODE_ENV`, bucket names, etc. |

### Order matters

`gcp-sa-json-base64` MUST run before `pem-private-keys-base64`. The GCP outer container holds a PEM-b64 substring; if we redact the inner first, the outer match no longer has enough remaining bytes to satisfy its length requirement and leaks.

## Backward compatibility

Render is identical at runtime. No env-var name, value, secret ref, or workload behavior changes. Verified with `helm template` against three scenarios:

- Fresh install / inline values — preflight runPod has 33 inline env vars, 0 refs → KOTS pre-install works
- `existingSecret` for postgres / redis / LD — render unchanged from main
- Disabled (no value, no existingSecret) — render unchanged from main

The only diff in any rendered output is the addition of `spec.redactors:` at the bottom of both preflight and support-bundle specs.

## Verification

- [x] `helm lint` passes
- [x] `helm template` renders clean across fresh-install + existingSecret + disabled scenarios
- [x] Preflight runPod podSpec has **0** `valueFrom` env entries — KOTS pre-install will not break
- [x] All 12 regex patterns compile as valid Go regexp (verified via Python re module which uses compatible syntax)
- [x] All shape-based patterns match their intended sentinels in a synthetic corpus
- [x] **Simulated redaction against `support-bundle-2025-08-25T13_54_03`**: PEM private keys (b64), GCP SA JSON (b64), and `eyAidHlwZSI6` GCP prefix all scrubbed. Bundle 48 949 → 30 273 bytes (18 KB of secrets removed).
- [x] **Simulated redaction against `replicated-license-info-stdout.txt`**: `-----BEGIN PRIVATE KEY-----`, `sk-svcacct-`, and `BEGIN PRIVATE` all scrubbed. 3 `***HIDDEN***` insertions.

The Azure storage key (`vA8/K5m+1i/...`) is shape-less and only covered by the scoped `yamlPath` fallback — my Python simulation doesn't model `yamlPath`, but real Troubleshoot does and will redact it. Worth confirming with `troubleshoot redact -f <rendered-spec> < real-bundle.tar.gz` once the binary is available in CI.

## Test plan

- [ ] CI lint + helm template pass
- [ ] CI gitleaks scan from #850 passes (no new secrets introduced)
- [ ] Reviewer renders chart locally with both `value:` and `existingSecret` flows and confirms `redactors:` block appears in both `preflight.yaml` and `support-bundle.yaml` Secrets
- [ ] Manual (recommended before merge): deploy to a test tenant, capture a support bundle with `kubectl support-bundle`, extract, `grep -rE 'BEGIN PRIVATE|AKIA|AIzaSy|sk-svcacct|sdk-[0-9a-f]{8}'` — should be zero hits

## Notes for reviewers

- The history on this branch is noisy — earlier commits introduced and then walked back a `secretKeyRef` refactor. Please **squash-merge** so `main` gets a single clean commit.
- Future credential shapes added to the platform must be added to the redactor set in `_commonChecks.tpl`. Add a regex for shapes, or extend the `carto-license-sensitive-entitlements` field list, or rely on the scoped `yamlPath` fallback for shape-less values.
- The `yamlPath` redactor in commit 3 (`tenant-requirements-check-env-fallback`) is intentionally scoped via `fileSelector` to a single file, addressing the Gemini reviewer's concern about over-redaction of debug bundles.

---

## Update 2026-07-06 — adversarial review findings folded in

An adversarial re-review with the real Troubleshoot binary (0.123.x) found and fixed two defects that the earlier claims above got wrong:

1. **The `tenant-requirements-check-env-fallback` rule never fired.** troubleshoot compiles `fileSelector` globs with gobwas/glob, where a `**/` prefix requires ≥1 parent directory — and `tenant-requirements-check/tenant-requirements-check.json` sits at bundle-root depth. Reproduced: the Azure storage key (the value this rule exists for) survived redaction. Fixed by adding the bundle-root-relative path to the fileSelector. The earlier claim "real Troubleshoot models yamlPath and will redact it" was wrong; it is now **verified by execution**, not assumed.
2. **The test couldn't catch #1.** The Azure key was in the fixture but missing from `SENTINELS`, and the preserved-values check asserted `LOG_LEVEL`/`NODE_ENV` survive *inside the fallback's own target file* — which only passed because the rule was dead. The sentinel is added (test now fails on any regression of the glob) and the preserved probes moved to a pod outside the fallback's scope, which is the correct "debug data elsewhere survives" assertion. Note this supersedes the earlier "debug bundles keep LOG_LEVEL, NODE_ENV" claim for the tenant-requirements-check file itself: every env value in that one file is masked by design; all other files keep their values.

Also folded in: merge of latest `main` (the new S3-compatible endpoint/forcePathStyle preflight env vars were audited — non-sensitive, no new redactor needed), pinned helm + SHA-pinned checkout with least-privilege permissions/concurrency/timeout on the workflow, gitleaks allowlist narrowed from `chart/tests/` to the single test script, and `tests/` added to `.helmignore` so credential-shaped sentinels never ship inside the packaged chart.

Current test output: **16/16 sentinels scrubbed (Azure key included), 0 leaks, out-of-scope debug values preserved 3/3.**

Known remaining gaps (follow-ups, unchanged from review): the in-cluster Redactor auto-discovery path (`--load-cluster-specs`) and the preflight redaction path have no e2e regression test (needs a kind cluster), and the license-entitlements regex is single-line — a pretty-printed license dump would defeat it (current SDK output is one entitlement per line).

https://claude.ai/code/session_012EmMYTPo4CPoi7uky9tS49


mateo-di:
## End-to-end verification — recap

Tested against a real GKE cluster (gke-infra / mateod ns / mateodi-sh customer values + secrets). With the redactor doc shipped as a standalone `kind: Redactor` in the same Secret, all 15/15 known plaintext shapes are scrubbed from the captured bundle, 18 `***HIDDEN***` insertions, non-sensitive env preserved. Negative control on `main` leaks the LD SDK key (the regression this PR fixes).

CI now exercises the redactor end-to-end via `chart/tests/test-redactors.sh` (one step in the new `chart-tests` workflow, using the reusable `.github/actions/chart-setup` composite). Future chart tests land as one more step in the same job.

Follow-up tracked in [sc-557085](https://app.shortcut.com/cartoteam/story/557085) for the upstream Replicated SDK license-info leak surface (chart-level regex covers it today; SDK-side redaction is the cleaner long-term fix).

mateo-di:
### Redesign: redact by context instead of cataloguing credential shapes

Merged `main` (42 behind — it moved `_commonChecks.tpl` by +83/-23 with the AI-proxy-DB check, bucket-CORS analyzers, `trustedCACerts` refactor and the disconnected auth/accounts stack), then replaced the regex catalog.

**13 rules → 3.** The old set was 11 credential-shape regexes (LaunchDarkly `sdk-`/`mob-`, AWS `AKIA`/`ASIA`, OpenAI `sk-`, Google `AIza`, JWTs raw + base64, PEM raw + base64, GCP SA JSON). That is a second secret inventory living in the chart, and it only covers the shapes someone remembered to add. Replaced with structured, file-scoped redaction:

| rule | scope | what it masks |
|---|---|---|
| `replicated-license-entitlement-values` | `**/replicated-license-info-std{out,err}.txt` | `entitlements.*.value` |
| `tenant-requirements-check-env-values` | `tenant-requirements-check/tenant-requirements-check.json` (+ `**/` form) | `spec.containers.*.env.*.value`, `spec.initContainers.*.env.*.value` |
| `api-key-json-fields` | all files | `"apiKey":"…"` in collected logs, plain and escaped |

Strictly better coverage: the fixture now includes an **unregistered secret format** that the old catalog would have missed and this set scrubs. `25/25` sentinels scrubbed, `0` leaks, `9/9` non-sensitive preserved. Also applied the remaining review nit (`| quote` on the postgres secret key).

### Verified on a live install

Real GKE cluster with Gatekeeper enforcing, dev test-customer package, 14/14 pods healthy, real `kubectl support_bundle` capture against the deployed spec.

**Deployed spec is correct** — a standalone `kind: Redactor` alongside `kind: SupportBundle`, 3 rules, both structured rules `fileSelector`-scoped.

**Scoping holds — this settles the "over-redaction is cluster-wide" concern.** Across *all* pod snapshots in `cluster-resources`, only 1 value was masked (by a Troubleshoot built-in) and non-sensitive env was kept in clear, e.g. `WORKSPACE_POSTGRES_SSL_MODE = disable`. The env-value rule does **not** bleed into other PodSpecs.

**Leak audit** — grepped the captured bundle for the actual values of 11 secrets from the customer package plus the GCP SA `private_key` body:

```
launchDarklySdkKey CLEAN   geminiApiKey CLEAN   vitallyToken CLEAN   jwtApiSecret CLEAN
encryptionSecretKey CLEAN  redisPassword CLEAN  varnish{Purge,Debug}Secret CLEAN
litellm{Master,Salt}Key CLEAN  gitbookApiToken CLEAN  GCP SA private_key CLEAN
raw PEM markers: 0 files   ***HIDDEN*** insertions: 36
```

The LaunchDarkly key appears **2× in the rendered spec** (inline by design — KOTS runs preflights pre-install, before chart Secrets exist) and **0× in the shipped bundle**. That is the mechanism working as intended.

**The checker log is untouched.** All check results and `.info` messages are readable, 0 redactions — so the primary diagnostic artifact is unaffected.

### Two things reviewers should weigh

**1. The env-value rule does mask non-sensitive values inside the checker's own snapshot.** 29 of 29 literal env values there become `***HIDDEN***`, including `WORKSPACE_POSTGRES_PORT` (`5432` on `main`) and `WORKSPACE_THUMBNAILS_PROVIDER` (`gcp`). Gemini's *scope* claim was wrong, but the *substance* is real and I do not want to paper over it: the trade-off is deliberate (the checker inlines shapeless secrets with no stable field or value shape) and bounded to one file, and the checker log above covers most debugging needs. Still a conscious call, not a free win.

**2. This Helm-path capture cannot demonstrate the leak it fixes, and that is informative.** On `main` the same audit also came back clean, because in this configuration the secrets that are set are already handled by Troubleshoot's built-ins (`*KEY`, `*TOKEN`, `*PASSWORD` env names), while the shapes this PR uniquely targets are not populated:

- `replicated-license-info-*` — a Replicated SDK artifact that only exists in a **KOTS** install. This is where the documented `cartoPlatformDefaultSA` / `openAiApiKey` / `vitallyToken` / entitlement leaks live.
- `DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT` / `ROUTER_SSL_CERT_KEY__FILE_CONTENT` — empty unless the install uses SA-key auth or a custom router cert.

So the security delta is **KOTS-path specific**, consistent with the original write-up being derived from real customer bundles. A KOTS install is required to demonstrate it, including the pre-install preflight path that the earlier `secretKeyRef` revision broke. That run is pending a free dedicated slot (all are currently reserved).

Marking this out of draft is gated on that KOTS run.

mateo-di:
### Correction, with the leak now demonstrated on the Helm path

My previous comment concluded the security delta was KOTS-only and could not be shown on a standalone Helm install. **That was wrong, and it understated this PR.** Two mistakes in the audit:

1. I resolved the customer secrets as `.cartoSecrets.<key>` (a map) instead of `.cartoSecrets.<key>.value`, so most of those "CLEAN" results were comparing against YAML map serialisations rather than real secret values — vacuous.
2. I probed the GCP service account with raw-PEM needles. The value reaches the checker as `DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT`, which is **base64** of the SA JSON — so no `-----BEGIN PRIVATE KEY-----` marker and no raw `private_key` body ever appears, and every raw needle came back clean on both sides.

Re-ran it with correctly resolved values and a base64 needle, same live install, same two bundles:

| needle | `main` (no redactors) | this PR |
|---|---|---|
| **GCP SA JSON, base64 slice (120 chars)** | **LEAKED** — `tenant-requirements-check/tenant-requirements-check.json` | **CLEAN** |
| SA `private_key` body (raw, 120 chars) | clean | clean |
| SA `private_key_id` | clean | clean |
| SA `client_email` | clean | clean |
| `geminiApiKey`, `vitallyToken`, `jwtApiSecret`, `encryptionSecretKey`, `redisPassword`, `litellmMasterKey` | clean | clean |

So there **is** a concrete, reproducible leak on the plain-Helm path, and this PR closes it: the full GCP service-account JSON sits base64-encoded in the captured checker snapshot on `main`.

It is worth being precise about why Troubleshoot's built-ins miss it, because it is the argument for this PR's whole approach:

- the env var name is `DEFAULT_SERVICE_ACCOUNT_KEY__FILE_CONTENT` — it does not match the built-in `*PASSWORD*` / `*TOKEN*` / `*ACCESS_KEY_ID` / `*SECRET_ACCESS_KEY` patterns,
- the value is base64, so no credential-shape regex for PEM blocks or JSON fields fires either.

A catalog of credential formats only catches this if someone adds a rule for "base64 of a GCP SA JSON" — which the previous revision of this PR did, with a rule matching three specific base64 prefixes of `{"type":"service_account"`. That is exactly the brittleness the redesign removes: the file-scoped `spec.containers.*.env.*.value` rule catches it because of *where* it is, with no knowledge of *what* it looks like.

The other secrets are clean on both sides because the built-ins already cover their env names — so this PR's marginal value is precisely the shapes the built-ins miss, and the SA JSON is a live example of one.

### What this does not change

The two caveats from my previous comment still stand:

- **Over-redaction inside the checker snapshot is real.** 29 of 29 literal env values there become `***HIDDEN***`, including `WORKSPACE_POSTGRES_PORT` and `WORKSPACE_THUMBNAILS_PROVIDER`. Bounded to that one file — other pod snapshots keep their env in clear (verified) — and the checker *log* with all check results is untouched. Still a deliberate trade-off worth a reviewer's opinion.
- **A KOTS run is still needed** for the `replicated-license-info-*` entitlements rule (a Replicated SDK artifact that does not exist in a Helm install) and for the pre-install preflight path the earlier `secretKeyRef` revision broke. That is queued on a dedicated environment.

mateo-di:
### KOTS validation done — and the leak is much worse than this PR's description implies

Brought up a dedicated-k8s environment on the released `Dedicateds` channel, then moved its dev-licensed customer to this branch's channel and upgraded the live install (`kots upstream upgrade`, sequence 2 -> 4). Confirmed the deployed `carto-support-bundle` Secret carries the standalone `kind: Redactor` doc with the 3 rules.

Then captured bundles two ways to isolate the effect.

**Method note, because it invalidated my first attempt:** Troubleshoot redactors apply to the *whole* bundle, not just the spec that contributed them. So a `--load-cluster-specs` run has this PR's redactors masking every namespace's files, and a namespace still on the released chart is **not** a valid control. The real control is to collect with the **Replicated SDK spec alone**, so no chart redactor is present at all.

**Control run — SDK spec only, no chart redactors.** `replicated-sdk/<ns>/<pod>/replicated-license-info-stdout.txt` is a single-line JSON blob (which is why field-by-field greps come back empty) containing, in **plaintext**:

| entitlement | what leaks |
|---|---|
| `cartoPlatformDefaultSA` | the **entire GCP service-account JSON**, including the full `-----BEGIN PRIVATE KEY-----` RSA body, `private_key_id`, and `client_email` |
| `ldsConfiguration` | **four** provider API keys — Google, HERE, Mapbox, TomTom |
| `openAiApiKey` | full `sk-svcacct-…` key |
| `geminiApiKey` | full `AIza…` key |
| `databaseEncryptionKey` | 32-char key |
| `jwtEncryptionKey` | 16-char key |
| `vitallyToken` | base64-wrapped JWT |
| `instanceId`, `cartoAuthClientId`, `selfHostedId` | identifiers |

`***HIDDEN***` count in that file: **0**.

**With this branch deployed:** every `entitlements.*.value` is `'***HIDDEN***'`, while `title`, `valueType` and `signature` are preserved so the artifact stays useful. `***HIDDEN***` count in that file: **16**.

### Why this raises the priority

The `replicated-license-entitlement-values` rule is not covering an edge case. **Every support bundle from every KOTS/Replicated customer today contains that customer's full GCP service-account private key, their LDS provider API keys, their OpenAI and Gemini keys, and their database and JWT encryption keys, in plaintext.** Troubleshoot's built-ins do not touch it: the leak is in a `.txt` artifact produced by the Replicated SDK's `exec` collector, the field is named `value` (not `*PASSWORD*`/`*TOKEN*`/`*KEY`), and the SA JSON's newlines are escaped so PEM-block regexes do not match either.

That also settles the design argument in this PR. A catalog of credential shapes only catches this if someone writes a rule per provider — and it would still miss `ldsConfiguration`, which is a JSON blob of four different vendors' key formats. The file-scoped `entitlements.*.value` rule catches all of it because of *where* it is.

I did not put any of the leaked values in this comment, and every bundle captured during this work has been destroyed. Worth noting separately: the credentials for this test environment (`ded-k8s-25`) were exposed to my local terminal during the control run, so they should be treated as needing rotation — they are shared dedicated-env credentials, not a customer's.

### Coverage summary

| quadrant | status |
|---|---|
| Helm, new install | done — GCP SA base64 in the checker snapshot: leaked on `main`, clean here |
| Helm, upgrade | done — 14/14 pods healthy through the upgrade |
| KOTS, upgrade from released | **done** — the entitlements leak above |
| KOTS, fresh install (pre-install preflight) | remaining — this is the path the earlier `secretKeyRef` revision broke |

The one open design question from my earlier comment still stands: inside the checker's own snapshot, 29 of 29 literal env values become `***HIDDEN***`, including non-sensitive ones like `WORKSPACE_POSTGRES_PORT`. Scoped to that one file, and the checker log with all check results is untouched — but it is a deliberate trade-off and I would like a second opinion on it rather than merging it silently.

mateo-di:
### The pre-install preflight concern, settled structurally

My earlier comments listed "KOTS fresh install (pre-install preflight)" as the last open quadrant — the path the abandoned `secretKeyRef` revision broke, where the runPod referenced chart-generated Secrets that do not exist yet, hit `CreateContainerConfigError`, sat Pending for the 180s timeout, and made KOTS block the deploy.

That concern does not apply to this revision, and it is provable from the diff rather than needing a rebuild. Against `origin/main`:

```
chart/templates/_commonChecks.tpl   | 47 ++++-    two hunks only
chart/templates/preflight.yaml      | 11 ++++
chart/templates/support-bundle.yaml | 15 ++++-

secretKeyRef / valueFrom lines changed: 0
```

The two `_commonChecks.tpl` hunks are:

- `@@ -648,7 +648,7 @@` — a single line: the `| quote` fix on an **existing** `secretKeyRef` key.
- `@@ -725,3 +725,48 @@` — a pure append at end of file: the `carto.replicated.commonChecks.redactors` define.

So the checker runPod's podSpec is byte-identical to `main`'s. This PR adds a `kind: Redactor` document and nothing else; it cannot regress a path it does not touch. The failure mode of the earlier revision is structurally impossible here.

Corroborated on the live environment anyway — across this session's deploys KOTS ran preflights 12 times with **8 successful `run-pod/tenant-requirements-check` collections**, including on this branch's channel, and:

```
CreateContainerConfigError : 0
couldn't find key          : 0
```

The only collector errors were `pods "tenant-requirements-check" already exists`, from my own back-to-back re-deploys racing the previous preflight's pod teardown. Unrelated to this change and present regardless of chart version.

A genuinely fresh `kots install` would still be the belt-and-braces check, and I'm happy to run one if a reviewer wants it — but on the diff above I don't think it changes the answer.

### Final coverage

| quadrant | result |
|---|---|
| Helm, new install | GCP SA JSON (base64) leaks in the checker snapshot on `main`, clean here |
| Helm, upgrade | 14/14 pods healthy through the upgrade |
| KOTS, upgrade from released | **full license-entitlement leak** on the released chart, all values `***HIDDEN***` here |
| KOTS, pre-install preflight | podSpec unchanged vs `main` (0 `valueFrom` deltas); 8 successful checker collections, 0 config errors |
| CI | `chart-tests` green — render contract on both paths + 25/25 sentinels scrubbed, 0 leaks, 9/9 non-sensitive preserved |

### One open question for review

Inside the checker's own snapshot, 29 of 29 literal env values become `***HIDDEN***`, including non-sensitive ones like `WORKSPACE_POSTGRES_PORT` and `WORKSPACE_THUMBNAILS_PROVIDER`. It is scoped to that single file — other pod snapshots keep their env in clear, and the checker *log* with every check result and `.info` message is untouched — but it is a deliberate trade-off, taken because that pod inlines shapeless secrets (Azure storage keys) with no stable field or value shape to target. I'd like that judged explicitly rather than merged quietly.

mateo-di:
### Clarification on what "KOTS validated" meant here

Correcting my own wording. Earlier comments on this PR described KOTS coverage without qualifying how it was produced, and one of them listed the fresh-install quadrant as effectively settled. Precisely what was and wasn't done:

**Done —** a live dev-licensed dedicated-k8s environment, moved onto this PR's Replicated channel via `replicated customer update` and upgraded in place with `kubectl kots upstream upgrade`. That genuinely exercises the KOTS render, the chart swap, and the upgrade path on an existing install, and the results I posted from it stand.

**Not done —** two distinct things:

1. **A fresh `kots install` from this PR's channel.** The channel override is consumed only at the first install, and by the time the harness environment pointed at this PR's channel it already existed, so every subsequent run took the update path (`Config Change`), never `Online Install`. Confirmed from the harness branch's own history.
2. **An update driven by commit → CI build → ded deploy.** My upgrade was manual. The pipeline's own `deploy-step: update` (`kots set config --deploy`) plus the backend-ci image rebuild were never involved.

For this PR the argument that the fresh-install path is safe is structural rather than empirical, and I still think it holds: the diff changes **zero** `secretKeyRef`/`valueFrom` lines, the checker runPod podSpec is byte-identical to `main`, and the only `_commonChecks.tpl` hunks are a one-line `| quote` and a pure end-of-file append. A regression in a path this PR does not touch is not possible. That is reinforced by the empty-namespace test, where the current spec's checker pod completes while the abandoned `secretKeyRef` revision fails with `secret "carto-externalpostgresql" not found`.

But structural reasoning is not the same as having run it, and the distinction matters more here than elsewhere because this PR's whole subject is what happens at collect time. The blocker is environmental — the dedicated pool is 35/35 reserved, so there is no slot for a fresh install. I'll run it as soon as one frees, and I'd rather this stayed in draft until then, consistent with the open design question about masking non-sensitive env values in the checker snapshot.

mateo-di:
### Stages 1 and 2 now done — fresh KOTS install from this branch's channel

The gap I flagged earlier is closed. This needed a dedicated environment installed **from scratch** against this PR's channel, which required its own throwaway harness (cloud-native#28726) and a free slot. Slot 031 was a leaked reservation from a closed PR; reclaimed it by triggering the existing `Dedicateds Housekeeping` workflow rather than hand-editing state.

**Stage 1 — fresh `kubectl kots install` from this branch's channel.**

The override was consumed at first install, which is the whole point (it is ignored on every later run):

```
Found channel to override: dedicated-selfhosted/replicated/override-channel/channel.yaml
replicatedChannel: "chore_preflight_secret_handling"
Updating ded-k8s-31 customer to chore_preflight_secret_handling channel
  -> channelName: chore_preflight_secret_handling
  -> channelSlug: chore-preflight-secret-handling
kubectl kots install "carto/${KOTS_CHANNEL_SLUG}"
```

KOTS records it as a genuine first install, not a config change:

```
VERSION    SEQUENCE    STATUS      SOURCE
1.295.1    0           deployed    Online Install
```

**The pre-install preflight — the thing that actually mattered.** This is the path the abandoned `secretKeyRef` revision broke, where the runPod referenced chart-generated Secrets that do not exist yet, hit `CreateContainerConfigError`, sat Pending through the 180s timeout, and made KOTS block the deploy. On this fresh install:

```
18:37:45  preflight progress: run-pod/tenant-requirements-check   status: running
18:38:23  preflight progress: run-pod/tenant-requirements-check   status: completed
18:38:23  preflight checks completed
```

`CreateContainerConfigError`: **0**. `couldn't find key`: **0**. The checker pod started and completed, and the install proceeded to deploy. So the argument I previously made structurally — that this PR cannot regress the pre-install path because it changes zero `secretKeyRef`/`valueFrom` lines — is now confirmed empirically.

And the environment really is running this branch's chart, not the released one — the deployed `carto-support-bundle` Secret carries the standalone `kind: Redactor` document:

```
SupportBundle
---
Redactor
  api-key-json-fields
  replicated-license-entitlement-values
  tenant-requirements-check-env-values
```

**Stage 2 — update over the existing install, driven by commit → CI build → KOTS deploy.**

My earlier KOTS check was a manual `replicated customer update` + `kots upstream upgrade`, which I called out as not being the pipeline. Pushed a commit to the harness branch while it held the environment:

```
CI - Backend / Deploy Selfhosted Dedicated : success
KOTS sequence 3 : deployed (Config Change)
```

and the branch's own images landed on the environment, proving the build reached it rather than redeploying stale tags:

```
carto-import-api      import-api:sc_854_kots_freshinstall_harness
carto-import-worker   import-api:sc_854_kots_freshinstall_harness
carto-maps-api        maps-api:sc_854_kots_freshinstall_harness
carto-sql-worker      maps-api:sc_854_kots_freshinstall_harness
```

Environment healthy throughout (32 Running, remainder rolling).

### Full stage coverage for this PR

| stage | status |
|---|---|
| 1. Fresh `dedicated-k8s` install from this PR's channel | **done** — `Online Install`, pre-install preflights completed, redactors deployed |
| 2. Update over existing install via commit → CI → KOTS deploy | **done** — `Deploy Selfhosted Dedicated: success`, sequence 3, branch images |
| 3. Preflights in a non-deployed namespace | done — checker pod completes; abandoned revision fails with `secret "carto-externalpostgresql" not found` |
| 4. Upgrade of an existing Helm deployment | done — 14/14 pods through the upgrade |

Plus the earlier KOTS-upgrade run that demonstrated the license-entitlement leak, and CI `chart-tests` (25/25 sentinels scrubbed, 0 leaks, 9/9 non-sensitive preserved).

Testing is complete. This is still in draft on purpose, for the one open design question: inside the checker's own snapshot, 29 of 29 literal env values become `***HIDDEN***`, including non-sensitive ones like `WORKSPACE_POSTGRES_PORT`. Bounded to that single file, other pod snapshots keep their env, and the checker log with all check results is untouched — but I would like that judged explicitly before this merges.

mateo-di:
### Out of draft — ready for review

Synced with `main` (picked up the `filetype=helm` modeline commit, which touches `_commonChecks.tpl`; no interaction with the end-of-file append here). CI green on the merged state: `build`, `chart-tests`, `Scan for secrets`, `check-helm-resources-changed`, `helm-readme-generator`, `release-changes`.

This thread is long, so a short orientation for whoever picks it up.

**What the diff is.** Three file-scoped Troubleshoot redactor rules appended to `_commonChecks.tpl`, included from `preflight.yaml` and `support-bundle.yaml`, plus two CI test scripts. No env var, secret ref, or workload behaviour changes — the checker runPod podSpec is byte-identical to `main` (0 `valueFrom` deltas), which is what keeps the KOTS pre-install preflight path safe.

**The leak it closes, in one line.** Every support bundle from every KOTS customer today contains that customer's full GCP service-account private key, their four LDS provider API keys, and their OpenAI/Gemini/database/JWT encryption keys, in plaintext, in `replicated-license-info-stdout.txt`. Troubleshoot's built-ins do not touch it: the field is named `value`, and the SA JSON's newlines are escaped so PEM regexes miss it. Demonstrated by control run, details above.

**Testing.** All four quadrants are done and evidenced in the comments — Helm fresh install (GCP SA base64 leaks on `main`, clean here), Helm upgrade, KOTS upgrade-from-released, and KOTS fresh install from this branch's channel with the pre-install preflight completing (`CreateContainerConfigError`: 0). Plus the pipeline path end-to-end (`Deploy Selfhosted Dedicated: success`).

### The one thing I want an explicit opinion on

Inside `tenant-requirements-check.json` — the checker's own snapshot — the `spec.containers.*.env.*.value` rule masks **29 of 29** literal env values, including non-sensitive ones like `WORKSPACE_POSTGRES_PORT` (`5432`) and `WORKSPACE_THUMBNAILS_PROVIDER` (`gcp`).

Why I took it anyway: that pod inlines shapeless secrets (Azure storage keys, base64 SA JSON) with no stable field name or value shape to target, so anything narrower is a credential catalog that only catches what someone remembered to add — which is exactly how the SA JSON leak survived the built-ins. It is bounded to that one file (verified: other pod snapshots keep their env in clear, e.g. `WORKSPACE_POSTGRES_SSL_MODE = disable`), and the checker *log* with every check result and `.info` message is untouched, so the primary diagnostic artifact is unaffected.

The alternative would be an allowlist of non-sensitive env-var name patterns (`*_PORT`, `*_PROVIDER`, `*_SSL_MODE`, …) so those stay readable. More moving parts, and it fails open by default — a new shapeless secret not on the denylist leaks until someone notices.

I'd rather that trade-off were judged than merged quietly. Happy to add the allowlist if the loss of those values in that one file is considered too costly.

### Merge note

25+ commits, with an abandoned `secretKeyRef` revision in the history. Please **squash-merge**.

PR #887, opened 2026-06-30
feat(chart): per-bucket S3-compatible endpoint/externalUrl/forcePathStyle values [sc-555243]

ManuelLR:
**Description of the change**

Adds optional S3-compatible storage configuration to the chart so CARTO Self-Hosted can point its main object-storage buckets at any S3-compatible provider (MinIO, Ceph, OVH, …) instead of AWS S3, GCS or Azure Blob.

Three new `appConfigValues`, all default-off:

- `storageEndpoint` — custom S3 endpoint for the main buckets
- `storageExternalUrl` — browser-facing external URL when it differs from the internal endpoint (split-horizon)
- `storageForcePathStyle` — path-style addressing, required by most S3-compatible providers

They are wired into the existing `storageProvider: "s3"` block of the `workspace-api`, `import-api` and `import-worker` configmaps, for both the thumbnails and imports bucket families. Each variable is only emitted when its value is non-empty, so a default install renders nothing new.

The imports family is exposed under both env-var namings the services use today — `WORKSPACE_IMPORTS_*` in `workspace-api` and `IMPORT_*` in `import-api`/`import-worker` — so the chart is already wired when imports gains S3-compatible support, even though the application does not read these variables yet.

**Benefits**

Operators can configure an S3-compatible backend through standard chart values, with the same shape as the existing S3 config — no custom or limited install. Defaults are unchanged, so existing AWS S3 / GCS / Azure deployments are unaffected.

**Possible drawbacks**

The `IMPORT_*` endpoint variables are emitted but not yet consumed by `import-api`/`import-worker` (imports S3-compatible support is a separate, deferred change); they are inert until that lands. Markers and branding inherit the thumbnails-family endpoint, matching the application's cascade. Running buckets private on providers without public-read ACL support is out of scope here and handled separately.

**Additional information**

Part of the "Self-Hosted support for any S3-compatible storage bucket" initiative (`sc-555243`).

Validated with `helm template` — default-off renders no new variables; with `storageProvider=s3` and the three values set, the new variables render in each service. `helm lint` reports no new errors (the pre-existing `engine: gotpl` warning is unrelated).

Opened as a draft: the feature ships default-off and pairs with a follow-up that surfaces these as KOTS Admin Console fields (`sc-555244`).


mateo-di (dismissed):
LGTM! 👏 Only minor thing I will add a warning in _validators when setting the new values when the storageProvider value is not s3

ManuelLR:
Thanks @mateo-di! Added in 25dcf4f — a `carto.validateValues.s3Compatible` check that fails the render (with a clear message) when `s3Endpoint`/`s3ExternalUrl`/`s3ForcePathStyle` are set but `storageProvider` isn't `s3`, consistent with the other `_validators` (hard fail rather than a soft warning, since that's what the file supports). Verified: default renders fine, misconfig fails, `storageProvider=s3` + endpoint renders fine.

mateo-di (approved):

PR #890, opened 2026-07-03
feat(chart): add components for selfhosted disconnected [sc-550114]

mateo-di:
**Description of the change**

Adds CARTO Self-Hosted **disconnected mode** to the chart: the in-cluster authentication service
(`auth-api`), the accounts stack (`accounts-api` + `accounts-subscriber`) it depends on, and the
wiring that switches every backend service and both SPAs to internal authentication. In disconnected
mode login is federated to a customer-controlled OIDC or SAML identity provider instead of the hosted
Auth0 tenant.

Strictly opt-in behind `appConfigValues.disconnectedEnabled` (default `false`). With the toggle off
the rendered chart is byte-identical to `main`. Templates never read the raw value; they read the
`carto.disconnected.enabled` helper, so the public toggle can evolve without touching templates.

What is included:

- **`auth-api` component** (`chart/templates/auth-api/`): configmap, deployment, service, hpa, pdb,
  secret. Follows the `aiApi` component pattern (`diagnosticMode`, `onlyRunRouter`, checksum
  annotations, external-proxy support).
- **Accounts stack** (`chart/templates/accounts-api/`, `chart/templates/accounts-subscriber/`).
  auth-api depends on accounts-api for M2M quota and SSO group sync, so it ships in the same PR.
- **Database wiring.** The accounts model and the internal-auth schema live in their **own database on
  the main chart PostgreSQL**: same host, same credentials, separate database named by
  `externalPostgresql.accountsDatabaseName` (default `"accounts"`). This follows the existing
  `aiProxyDatabaseName` idiom, and like that one the database must be created beforehand. Never a
  separate host: an internal PostgreSQL is a single instance, and a managed external one gets a new
  database from the DBA, not a new instance.
- **Optional dedicated PostgreSQL user for auth-api** (`authApi.postgresql.user` +
  `authApi.postgresql.password`, default empty). auth-api owns the `auth` schema, so it can use a
  login role scoped to that schema instead of the shared platform user. Only user and password
  diverge; host, port and database stay shared. Consume-only: the DBA pre-creates the role and its
  grants. Applies to the auth-api container and its migrations initContainer.
- **Shared env helpers.** `carto.disconnected.commonEnv` injects the internal-auth vars into the
  backend configmaps; `carto.disconnected.wwwEnv` points both SPAs at the internal issuer. Both are
  self-gated, so consumers include them unconditionally and never branch on the toggle themselves.
  The SPA authority and the token issuer come from one helper, since a mismatch fails the OIDC code
  flow.
- **Validators** (`_validators.tpl`): with the toggle on, missing required values fail
  `helm install` early with per-field messages.
- **Router**: conditional `ROUTER_ENABLE_AUTH_API`, `ROUTER_AUTH_API_INTERNAL_URL` and
  `ROUTER_ACCOUNTS_API_INTERNAL_URL`.
- **Secrets**: reuses the existing `cartoSecrets.encryptionSecretKey`; new
  `authApi.internalServiceToken` (`value` / `existingSecret`) shared between the auth-api and
  accounts-api internal endpoints; the OIDC client secret accepts an inline value or an
  `existingSecret`.
- **KOTS manifests**: new "Internal Authentication (Self-Hosted Disconnected)" config group, values
  mapping, conditional statusInformer.
- **README** regenerated with readme-generator-for-helm.

**Benefits**

- Self-Hosted installations can authenticate against their own identity provider, with no dependency
  on a hosted Auth0 tenant, which is the prerequisite for running fully disconnected.
- Existing installations are unaffected until they opt in: the default render does not change, and no
  new pod, migration or environment variable appears.
- Credential hardening is available without a topology change, through the optional dedicated
  PostgreSQL user for the service that owns the `auth` schema.
- Internal URLs (JWKS, auth-api, accounts-api) resolve through Services rather than pod ports,
  consistent with every other internal URL in the chart.

**Possible drawbacks**

- The accounts database must be pre-created on the platform PostgreSQL before enabling the toggle,
  exactly like the aiProxy database. An existence preflight is a candidate follow-up.
- Redis/Valkey becomes a runtime dependency of the login, refresh and logout flows, not just a cache.
  Session durability across restarts is handled separately in #911, which pairs with a change in the
  internal build pipeline and lands with it.
- Switching the auth-api PostgreSQL user after the first install is not zero-downtime, and full
  isolation also needs the DBA to revoke the shared user's access to the `auth` schema. Both are
  documented in `values.yaml`.
- Enabling the toggle adds three Deployments and two Services to the namespace.
- The chart carries no messaging-transport changes for disconnected mode. Event transport is
  configured through the generic `extraEnvVars` surface, so it stays out of the chart's contract.

**Applicable issues**

- Story: [sc-550114](https://app.shortcut.com/cartoteam/story/550114)
- Epic: [sc-544413](https://app.shortcut.com/cartoteam/epic/544413) (Self-Hosted Disconnected:
  Authentication System)

**Additional information**

**Open for review, not yet for merge.** This depends on routing and build-pipeline work tracked
separately in the internal repository and should not be merged or released until that lands. Do not
enable `appConfigValues.disconnectedEnabled` on any installation yet.

Validation:

- `helm lint chart/` → 0 failures.
- **Off-by-default proof**: `helm template` on `origin/main` and on this branch with identical inputs
  → `diff` is empty once render-time randomness (autogenerated TLS material, generated passwords) is
  pinned or excluded.
- Enabled render, OIDC: every auth-api, accounts-api and accounts-subscriber resource renders; the
  issuer and public base URL derive as `https://<selfHostedDomain>/auth-api`, callback URLs derive per
  protocol, secrets wire through env or secretRef, `ACCOUNTS_POSTGRES_DB` renders the accounts
  database name, and the router configmap carries `ROUTER_ENABLE_AUTH_API`.
- Enabled render, SAML variant: the SAML vars render, and a null `authApi.saml.metadataXml` no longer
  fails the render.
- Validator failure modes render clear per-field messages.
- `scripts/test-kots-config.sh`: all static tests pass.
- Exercised end to end on an internal Kubernetes test environment: install with the toggle off,
  enable it on the live environment, switch the identity protocol in both directions, and disable it
  again, checking at each step that the expected resources appear or disappear and that login works.

Review focus:

- The **off-by-default guarantee**: every changed shared file (`_helpers.tpl`, `_commonChecks.tpl`,
  `_validators.tpl`, the backend configmaps, the router configmap, the KOTS manifests) only adds
  toggle-gated blocks.
- The accounts-database model: a separate database on the main PostgreSQL via
  `externalPostgresql.accountsDatabaseName`, never a separate host.
- That the SPA authority and the token issuer are derived from the same helper.


chart/templates/accounts-api/configmap.yaml
```diff
@@ -0,0 +1,85 @@
...
+  ACCOUNTS_PARTNER_CONNECT_SNOWFLAKE_PARTNER_SECRET: "disabled"
+  BOBSLED_EXPORT_STORAGE_BUCKET: "disabled"
+  BOBSLED_SOURCE_ID: "disabled"
+  BOBSLED_ENVIRONMENT_NAME: "disabled"
+  BOBSLED_CLIENT_ID: "disabled"
+  BOBSLED_ACCOUNT_EMAIL: "disabled"
+  BOBSLED_ACCOUNT_SECRET: "disabled"
+  ACCOUNTS_BACKOFFICE_AUTH_USER: "disabled"
+  ACCOUNTS_BACKOFFICE_AUTH_PASSWORD: "disabled"
+  ACCOUNTS_ACTIVITY_DATA_BUCKET_NAME: "disabled"
```
mateo-di, line 73: check

chart/templates/accounts-api/configmap.yaml
```diff
@@ -0,0 +1,85 @@
...
+  # Defaults to true upstream and would attempt outbound Slack notifications
+  QUOTA_NOTIFICATIONS_ENABLED: "false"
```
mateo-di, line 75: 🤔

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,89 @@ spec:
           default: "0"
 
 
+    # Self-Hosted Disconnected
+    - name: disconnected
```
moicalcob: Let's hide this section under a license field as we just want to allow enabling the disconnected under some circumstances. With these changes all customers using KOTS will see this option in the next release ⚠️

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,89 @@ spec:
...
+      title: Self-Hosted Disconnected
+      description: |-
+        Run CARTO Self-Hosted in disconnected mode: authentication is served by the internal authentication service (auth-api), federated to your own OIDC or SAML identity provider, and the accounts stack runs inside the cluster.
```
moicalcob: Let's please check with product all this copies 🙏🏼 We can iterate after merging

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,89 @@ spec:
...
+          title: Allowed CORS origins
+          help_text: |-
+            Comma-separated list of origins allowed by auth-api CORS.
+          type: text
+          when: '{{repl (ConfigOptionEquals "disconnectedEnabled" "1") }}'
+          default: 'https://{{repl ConfigOption "cartoSelfHostedDomain" }}'
+        - name: authApiPostgresUser
+          title: Dedicated PostgreSQL user for auth-api (advanced)
+          help_text: |-
+            Optional hardening. A dedicated PostgreSQL login role for auth-api only, scoped to the `auth` schema (which holds the signing keys) inside the accounts database. Leave empty to reuse the shared platform user. The role and its grants must be created on the database beforehand — the chart only consumes the credentials. If you set this, you must also set the password below.
```
moicalcob: Okay for me to merge it like this but let's think about this before a private preview for the disconnected

manifests/kots-helm.yaml
```diff
@@ -126,6 +128,40 @@ spec:
...
+          value: '{{repl ConfigOption "authApiPostgresPassword" }}'
+      oidc:
+        issuerUrl: '{{repl ConfigOption "authApiOidcIssuerUrl" }}'
+        clientId: '{{repl ConfigOption "authApiOidcClientId" }}'
+        clientSecret: '{{repl ConfigOption "authApiOidcClientSecret" }}'
+      saml:
+        metadataUrl: '{{repl ConfigOption "authApiSamlMetadataUrl" }}'
+        metadataXml: |
+          repl{{- ConfigOption "authApiSamlMetadataXml" | nindent 10 }}
+    authApiMigrations:
```
moicalcob: I'd call it auth migrations instead 🤔 To not link it to auth api

chart/values.yaml
```diff
@@ -6541,3 +6547,1094 @@ aiProxy:
...
+  protocol: ""
+  ## @param authApi.publicBaseUrl External base URL where auth-api is exposed. If empty, `https://<appConfigValues.selfHostedDomain>/auth-api` is used
+  ##
+  publicBaseUrl: ""
+  ## @param authApi.issuer Issuer (`iss` claim) of the tokens minted by auth-api. If empty, the computed `authApi.publicBaseUrl` is used
+  ##
+  issuer: ""
+  ## @param authApi.audience Audience (`aud` claim) of the tokens minted by auth-api
+  ##
+  audience: "carto-cloud-native-api"
```
moicalcob: This should not be parameterized I think that it's not in other apis

chart/values.yaml
```diff
@@ -6541,3 +6547,1094 @@ aiProxy:
...
+  defaultNodeProcessMaxOldSpace: 768
+  ## @param authApi.protocol Protocol used to federate authentication to the external identity provider. Allowed values: `oidc` or `saml`
+  ##
+  protocol: ""
+  ## @param authApi.publicBaseUrl External base URL where auth-api is exposed. If empty, `https://<appConfigValues.selfHostedDomain>/auth-api` is used
+  ##
+  publicBaseUrl: ""
+  ## @param authApi.issuer Issuer (`iss` claim) of the tokens minted by auth-api. If empty, the computed `authApi.publicBaseUrl` is used
+  ##
+  issuer: ""
```
moicalcob: Same as below I think? let's check all this kind of params as we just want to expose what is needed

chart/values.yaml
```diff
@@ -6541,3 +6547,1094 @@ aiProxy:
...
+  ## @param authApi.audience Audience (`aud` claim) of the tokens minted by auth-api
+  ##
+  audience: "carto-cloud-native-api"
+  ## @param authApi.orgId Organization ID (`org_id` claim) of this self-hosted installation
+  ##
+  orgId: "org_carto_sh_local"
+  ## @param authApi.accountsApiUrl Base URL of the accounts-api service that auth-api uses for quota checks and SSO group sync
+  ##
+  accountsApiUrl: ""
```
moicalcob: We can derive it in the config yaml. Please let's check all these params 🙏🏼

chart/values.yaml
```diff
@@ -6541,3 +6547,1094 @@ aiProxy:
...
+    ## Secrets must be manually created in the namespace.
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+    ## e.g:
+    ## pullSecrets:
+    ##   - myRegistryKeySecretName
+    ##
+    pullSecrets: []
+  ## @param authApiMigrations.command Override default container command (useful when using custom images)
+  ##
+  command: []
```
moicalcob, line 6981: By default accounts-db image does not apply auth api migrations no? How is this working?

chart/templates/maps-api/configmap.yaml
```diff
@@ -15,6 +15,9 @@ data:
   AUTH0_AUDIENCE: "carto-cloud-native-api"
   AUTH0_DOMAIN: {{ .Values.cartoConfigValues.cartoAuth0CustomDomain | quote }}
   AUTH0_NAMESPACE: "http://app.carto.com"
```
moicalcob, line 18: I think that if we enable disconnected we should get rid of all Auth0 env vars just in case. This applies to all services

moicalcob (changes requested):
Almost there 💪🏼

manifests/kots-config.yaml
mateo-di: Fixed! Gated behind the `installerFeaturesEnabled` license field (`disconnectedEnabled` or `all`), same pattern as the custom encryption keys.

manifests/kots-config.yaml
mateo-di: Agreed, will review the copy with product and iterate after merging.

manifests/kots-config.yaml
mateo-di: Agreed, let's discuss the dedicated DB user model before the private preview.

manifests/kots-helm.yaml
mateo-di: Fixed! Renamed to `authMigrations`.

chart/values.yaml
mateo-di: Fixed! Hardcoded, matching the other APIs.

chart/values.yaml
mateo-di: Fixed! Always derived from the public base URL.

chart/values.yaml
mateo-di: Fixed! Always the in-cluster service now.

chart/values.yaml
mateo-di: The accounts-db image ships both `migrations/` and `migrations-internal-auth/`; the initContainer overrides the args to run the internal-auth set explicitly (`db-migrate up -m migrations-internal-auth --table internal_auth_migrations`).

chart/templates/maps-api/configmap.yaml
mateo-di: Fixed! All Auth0 env vars are omitted when disconnected mode is enabled.

chart/values.yaml
```diff
@@ -5594,6 +5597,9 @@ externalPostgresql:
   awsEksPodIdentityEnabled: false
   awsRdsRegion: ""
   aiProxyDatabaseName: "aiproxy"
+  ## The accounts database rides the main PostgreSQL on purpose: one instance, one service user,
+  ## separate database (industry-standard default — Keycloak/ZITADEL ship the same shape).
```
ManuelLR: Please, review all the comments on the PR. For example that doesn't makes sense

chart/values.yaml
```diff
@@ -5594,6 +5597,9 @@ externalPostgresql:
...
+  accountsDatabaseName: "accounts"
```
ManuelLR, line 5655: Are we sure we are going to transmit to the client from now on that inside the SH we have a component called account? Do we sit with the PMs to check it?

chart/values.yaml
```diff
@@ -6541,3 +6547,1085 @@ aiProxy:
...
+      ##
+      existingSecret:
+        name: ""
+        key: ""
+  ## @param authApi.accessTokenTtlSeconds Access token TTL in seconds. Leave empty for the service default
+  ##
+  accessTokenTtlSeconds: ""
+  ## @param authApi.refreshTokenTtlSeconds Refresh token TTL in seconds. Leave empty for the service default
+  ##
+  refreshTokenTtlSeconds: ""
```
ManuelLR: If I'm not wrong, we usually define all that things in other place and we let all the things inside the *Api to pure container/pod/deployment definition.

chart/values.yaml
```diff
@@ -6541,3 +6547,1085 @@ aiProxy:
...
+  ##
+  ##
+  service:
+    ## @param authApi.service.type auth-api service type
+    ##
+    type: ClusterIP
+    ## @param authApi.service.ports.http auth-api service HTTP port
+    ##
+    ports:
+      http: 80
```
ManuelLR, line 6918: Are we  sure that's the port? we usually assign the same that is locally available and we avoi that kind of reserved ports that could cause problems in some deployments with more extrict rules about ports.

chart/values.yaml
```diff
@@ -6541,3 +6547,1085 @@ aiProxy:
...
+    ## @param authApi.service.annotations Additional custom annotations for auth-api service
+    ##
+    annotations: {}
+    ## @param authApi.service.extraPorts Extra ports to expose in auth-api service (normally used with the `sidecars` value)
+    ##
+    extraPorts: []
+
+## @section auth Migrations Init Container Parameters
+##
+authMigrations:
```
ManuelLR, line 6957: Why call it authMigrations instead of accountsMigrations? 🤔

ManuelLR: Ok, I found that we have also accountsMigrations. Do we really create another migrations using the same container but they don't apply at the same time? 🙈

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,91 @@ spec:
           default: "0"
 
 
+    # Self-Hosted Disconnected
+    # Hidden unless the license enables it: the feature is not generally available yet.
+    - name: disconnected
+      title: Self-Hosted Disconnected
+      when: '{{repl or (eq (LicenseFieldValue "installerFeaturesEnabled") "all") (contains "disconnectedEnabled" (LicenseFieldValue "installerFeaturesEnabled")) }}'
```
ManuelLR: Note for me, in every config in the HELM config that's related with the disconnected, we need to check the value (of course) and the license field. That second one is required if we're going to hide things in that way (and it's normal)

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,91 @@ spec:
...
+      description: |-
+        Run CARTO Self-Hosted in disconnected mode: authentication is served by the internal authentication service (auth-api), federated to your own OIDC or SAML identity provider, and the accounts stack runs inside the cluster.
+      items:
+        - name: disconnectedEnabled
+          title: Enable disconnected mode
+          help_text: |-
+            Deploy the internal authentication service (auth-api) and the in-cluster accounts stack, and switch every backend service to internal authentication.
+          type: bool
+          default: "0"
```
ManuelLR: If this is going to be something in the license, we should enforce it.

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,91 @@ spec:
...
+          title: SAML identity provider metadata XML
+          help_text: |-
+            Inline metadata XML. Preferred in production (no boot-time network dependency).
+          type: textarea
+          when: '{{repl and (ConfigOptionEquals "disconnectedEnabled" "1") (ConfigOptionEquals "authApiProtocol" "saml") }}'
+        - name: authApiOrgId
+          title: Organization ID
+          type: text
+          when: '{{repl (ConfigOptionEquals "disconnectedEnabled" "1") }}'
+          default: "org_carto_sh_local"
```
ManuelLR: I don't know what to say about that. Would we really like it to be a free field instead of reading it from the license or deriving it from the license? Would we really like to use the default value? Would a client understand what that simple title entails without a description?

Will we allow clients to change the field after it has been defined?

manifests/kots-config.yaml
```diff
@@ -1032,6 +1042,91 @@ spec:
...
+          readonly: false
+          when: '{{repl (ConfigOptionEquals "disconnectedEnabled" "1") }}'
+          value: '{{repl RandomString 32 }}'
+        - name: authApiAllowedOrigins
+          title: Allowed CORS origins
+          help_text: |-
+            Comma-separated list of origins allowed by auth-api CORS.
+          type: text
+          when: '{{repl (ConfigOptionEquals "disconnectedEnabled" "1") }}'
+          default: 'https://{{repl ConfigOption "cartoSelfHostedDomain" }}'
```
ManuelLR: Are we going to allow define custom CORS headers for that API but not for others? That's strange, isn't it?

manifests/kots-helm.yaml
```diff
@@ -85,6 +85,7 @@ spec:
     appConfigValues:
       aiFeaturesEnabled: repl{{ if ConfigOptionEquals "aiFeaturesEnabled" "1"}}truerepl{{ else }}falserepl{{ end }}
       publicEventsApiEnabled: repl{{ if ConfigOptionEquals "publicEventsApiEnabled" "1"}}truerepl{{ else }}falserepl{{ end }}
+      disconnectedEnabled: repl{{ if ConfigOptionEquals "disconnectedEnabled" "1"}}truerepl{{ else }}falserepl{{ end }}
```
ManuelLR: I think all the disconnected things should be applied as other layer instead of adding it to the default one. That's going to simplify the understanding and let us confirm that the helm keep working as expected without that values :)

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
       value: "True"
 {{- end -}}
+
+{{/*
+Whether this deployment runs in self-hosted disconnected mode. Every disconnected gate must read
+this helper instead of the raw value, so the public toggle can evolve without touching templates.
+Returns "true" when enabled, empty string (falsy) otherwise.
+*/}}
+{{- define "carto.disconnected.enabled" -}}
+{{- if .Values.appConfigValues.disconnectedEnabled -}}true{{- end -}}
```
ManuelLR: Ey, it's not so easy. We should check also the license value. Remember the hidden or not shown fields preserves the value previously defined.

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
+
+{{- define "carto.authMigrations.image" -}}
+{{- include "carto.images.image" (dict "imageRoot" .Values.authMigrations.image "global" .Values.global "Chart" .Chart) -}}
+{{- end -}}
+
+{{- define "carto.authApi.publicBaseUrl" -}}
+{{- if .Values.authApi.publicBaseUrl -}}
+{{- trimSuffix "/" .Values.authApi.publicBaseUrl -}}
+{{- else -}}
+{{- printf "https://%s/auth-api" .Values.appConfigValues.selfHostedDomain -}}
```
ManuelLR: That's so hilarious. Do you know we aren't routing to the auth-api if the path doesn't end with `/` ?

I think something is wrong in some place. Maybe here, maybe the auth-api code or maybe the router-http config.

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Client ID of the boot-seeded platform SPA client. auth-api registers it and both SPAs
+authenticate with it, so all three must read this one helper or login breaks on a mismatch.
+*/}}
+{{- define "carto.authApi.spaClient.clientId" -}}
+{{- default "carto-spa" .Values.authApi.spaClient.clientId -}}
+{{- end -}}
```
ManuelLR: I think we shouldn't add a default and it should fails instead. That should be defined always except for testing right? 🤔

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
+Redirect URIs registered on the platform SPA client. The default covers both SPAs: workspace-www
+redirects to the bare origin, accounts-www to /callback. Registering only one 400s the other.
+*/}}
+{{- define "carto.authApi.spaClient.redirectUris" -}}
+{{- if .Values.authApi.spaClient.redirectUris -}}
+{{- join "," .Values.authApi.spaClient.redirectUris -}}
+{{- else -}}
+{{- printf "https://%s,https://%s/callback" .Values.appConfigValues.selfHostedDomain .Values.appConfigValues.selfHostedDomain -}}
+{{- end -}}
+{{- end -}}
```
ManuelLR: I think that's not the exact, right? do we redirect the callback to the workspace-www? (yes, the /callback is workspace-www)

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
+
+{{/*
+The disconnected-mode environment package for the SPAs. Self-gated like
+carto.disconnected.commonEnv, so consumers include it unconditionally. The authority must be the
+issuer auth-api mints with, hence the shared helper — a mismatch fails the OIDC code flow.
+*/}}
+{{- define "carto.disconnected.wwwEnv" -}}
+{{- if (include "carto.disconnected.enabled" .) -}}
+REACT_APP_AUTH_PROVIDER: "oidc"
```
ManuelLR, line 1729: Provider hardcoded ⚠️

chart/templates/_helpers.tpl
```diff
@@ -1549,3 +1550,205 @@ httpGet:
...
+REACT_APP_OIDC_AUTHORITY: {{ include "carto.authApi.issuer" . | quote }}
+REACT_APP_OIDC_CLIENT_ID: {{ include "carto.authApi.spaClient.clientId" . | quote }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Base URL of the in-cluster accounts-api service that auth-api calls for quota checks and SSO
+group sync.
+*/}}
+{{- define "carto.authApi.accountsApiUrl" -}}
```
ManuelLR: I think here we should call it internal or something like that, if not, it could cause us problems 🤔

ManuelLR (changes requested):

chart/values.yaml
mateo-di: That's a comment that Claude left related to how the accounts db public and auth schemas are separated in the same db. WIll remove it since it doesn't make sense to keep and has no context.

chart/values.yaml
mateo-di: Yes, I think that should now be included in the Disconnected initiative as part of the chart managed databases. We can check with PMs but I don't see a world were this is not specified in the chart as the rest of the database for the SH 🤔 do you have any idea for this?

chart/values.yaml
mateo-di: Yep for the service is normally that one 🤔

chart/values.yaml
mateo-di: I tried to keep per component migration initcontainers 🤔 i didn't want to tied auth-api deployability depend on accounts-api's container/deployment (or viceversa). You think we should run them in 1? what would be the benefit?

chart/values.yaml
mateo-di: You are totally right. My blunder. Fixed.

manifests/kots-config.yaml
mateo-di: This was an idea that I don't have fully clear now... I think is better to drop all Kots configuration in this PR and add it in a followup.

manifests/kots-config.yaml
mateo-di: same as above

manifests/kots-config.yaml
mateo-di: same as above

manifests/kots-config.yaml
mateo-di: same as above

manifests/kots-helm.yaml
mateo-di: same as above

chart/templates/_helpers.tpl
mateo-di: I think nothing is broken! This is a helpers "hack" for the Auth-api interaction with the IdP's. The router strips /auth-api before the request reaches the container, and auth-api is the one component that must hand out absolute self-referencing URLs to the IdP's outside the cluster. I had problems with this and added the helper. 

I investigated some altenratives and this looked like the simpler one: pinnin the config with an explicit external base url (saw that other sh idps do the same). But i think there are alternatives deriving it from the headers or maybe something else you have in mind?

chart/templates/_helpers.tpl
mateo-di: Agree, fixed.

chart/templates/_helpers.tpl
mateo-di: I think you're right 🤔 the comment (and the default) are stale, from what i checked none of the SPAs actually uses /callback 🤔  workspace-www logs in from the origin and accounts-www from its /acc/ app root, right? 🤔 . so /callback is a dead fallback. 

I'll fix the default and the comment to match that derivation

chart/templates/_helpers.tpl
mateo-di: From what i understand the SPA never talks SAML/OIDC to the upstream IdP directly, only a standard OAuth2 code+PKCE flow to auth-api's own /authorize. 

Whether auth-api redirects onward to an OIDC or SAML IdP is decided entirely server-side in the auth-api authorization routes. So from the SPA's side auth-api always looks like a plain OIDC provider, regardless of protocol 

(this disconnected `wwwEnv` are only for the frontends) 

So i think this should stay a constant no? 🤔

chart/values.yaml
mateo-di: Done in 419d657: everything that is not a container/pod definition moved out of authApi.* — functional settings into appConfigValues.disconnected.*, the OIDC secret into appSecrets, the internal token into cartoSecrets, the auth-api-only PG role into externalPostgresql next to the accounts database it scopes to. authApi.* is now image/pod/resources only, like every other component 🙂

chart/templates/_helpers.tpl
mateo-di: You were right, fixed in 419d657. Neither SPA actually uses /callback (dead fallback from before the www-auth bridge): workspace-www redirects to the bare origin, accounts-www to its /acc/ app root. The default now matches exactly what accounts-api's tenant-up sync writes at first boot, so the seed is correct even before that sync runs.

chart/values.yaml
ManuelLR: I'm not asking about hiding the "Accounts" name. I'm asking if calling that component "Accounts" is the better name or if it could cause confusion with other things or other ideals from the client point of view.

chart/values.yaml
ManuelLR: cc: @moicalcob do you remember if that could be conflictive in some rootless environment? I remember something but I'm not 100% sure to insist more here

chart/values.yaml
ManuelLR: Today we have one migration container for all the "tenant". I understand that having it in splitted containers could be good if they are different databases but are they? 

If they are different databases, does it really make sense to reuse the same container and mix things? :thinking: 

cc: @moicalcob

manifests/kots-config.yaml
ManuelLR: I'm going to be agree with you in order to stop and sit how are we going to proceed. I'm not sure if only the kots side or all :thinking: 

~~I think we can proceed with all in the same PR, don't worry. In fact, that was a placeholder comment.~~

~~You only need to improve how you inject the values. I think that having all in kots-helm.yaml as an option with a simple "when" could simplify all and then you can iterate with a base already defined :)~~

~~If you don't understand something or don't know how to proceed, we can arrange a meeting and do together without problems.~~

chart/templates/_helpers.tpl
ManuelLR: :roll_eyes:

chart/templates/_helpers.tpl
ManuelLR: I wasn't referring to the hack itself.

Now that I notice, we shouldn't expose the `.Values.appConfigValues.disconnected.publicBaseUrl` because it's going to be always the same, in fact, you can directly inject it on every required variable without ifs.

Now, going back to the original comment, I was saying that we define `https://pepito/auth-api` but if you use exactly that url in the browser the router is going to go to workspace-www, you need to use `https://pepito/auth-api/`. I checked and it's true that's always defined in that way for all our APIs :thinking:

chart/templates/_helpers.tpl
ManuelLR: Perfect. Another thing. That variable should be really exposed to our clients? is it going to be always the same and an internal variable, right? That could simplify things :)

chart/templates/_helpers.tpl
ManuelLR: Totally agree with you. Sorry for the noise :see_no_evil:

chart/values.yaml
moicalcob: Nothin that I can remember honestly 🤔

chart/values.yaml
moicalcob: Well that's a bit special 😂 For SH we'll now have this:

1. Tenant DB with the tenant migrator as we had before
2. Accounts DB with the accounts migrations that we also run in SaaS
3. Auth schema inside Accounts DB that is used by Auth API and completely isolated but migrated by a set of migrations inside accounts migrations that are gated

The reason why we did it like this is because we wanted to use the same DB to avoid asking for 3 databases to a SH customer 🙇🏼

chart/values.yaml
mateo-di: Not sure actually 🤔 which alternative do you see to the name?

manifests/kots-config.yaml
mateo-di: 👌

chart/templates/_helpers.tpl
mateo-di: Fixed! Dropped publicBaseUrl, it's derived directly now.

On the trailing slash: good to know it's consistent everywhere, nothing to change here then right? Anything caughting your attention?

chart/templates/_helpers.tpl
mateo-di: You're right, dropped it. It always has to match accounts-api's tenant-up derivation exactly, so there's no real case for a customer value to diverge. Fixed in 5d72007, same treatment as `publicBaseUrl` a commit ago: derived only in `_helpers.tpl`, no override in `values.yaml`.

chart/templates/_helpers.tpl
mateo-di: Not sure if im following this one? 🤔 hidden? you are talking about kots config? for now the disconnected won't be tied to any license value AFAIK.

chart/templates/_helpers.tpl
ManuelLR: All good :)

chart/values.yaml
mateo-di: We keep it as is for now. Todo followup to re-review the name with PMs.

moicalcob (approved):

ManuelLR (approved):

PR #874, opened 2026-05-28
Bump import-worker CPU limit from 2 to 4 vCPU [sc-554030]

ManuelLR:
## Summary

Raise `importWorker.resources.limits.cpu` in `chart/values.yaml` from `2000m` to `4000m` to keep parity with the SaaS bump performed in `cloud-native`. The `import-worker` is the Pub/Sub subscriber that processes uploaded files and becomes CPU-bound on larger payloads at 2 vCPU.

Story: [sc-554030](https://app.shortcut.com/cartoteam/story/554030)
Companion PR (SaaS / Terraform): https://github.com/CartoDB/cloud-native/pull/25064

## What changes

- `chart/values.yaml`: `importWorker.resources.limits.cpu` 2000m → 4000m
- `chart/README.md`: regenerated entry to reflect the new default
- Memory limit (`8192Mi`), `requests.memory` (`3072Mi`) and `requests.cpu` (`350m`) are **unchanged**.

## Deployment impact

- Default chart now schedules the `import-worker` pod with a 4 vCPU limit. The CPU request remains `350m`, so existing K8s clusters can still schedule the pod, but bursts will now be allowed up to 4 vCPU on a single node.
- Customers on tight node sizing may want to adjust their cluster autoscaler or node pool. Should be mentioned in the next Selfhosted release notes.

## Validation steps

- [ ] `helm template` shows the rendered `import-worker` Deployment with `resources.limits.cpu: "4000m"` and unchanged memory/requests.
- [ ] No drift between this chart and the SaaS module after both PRs are merged.
- [ ] On a test Selfhosted environment, run a sizeable import and verify the new CPU ceiling is being used.

PR #892, opened 2026-07-05
feat(chart): size import-worker for the DuckDB import engine

mateo-di:
**Description of the change**

The import-worker now runs the DuckDB import/export engine alongside the legacy GDAL pipeline, but the chart still ships pre-DuckDB sizing and no DuckDB configuration. DuckDB allocates native memory outside the Node heap, bounded by `TRANSFER_DUCKDB_MEMORY_LIMIT` — unset, it defaults to 4 GB while the heap takes ~4 GiB of the same 8 GiB limit, leaving no headroom. Mirrors CartoDB/cloud-native#25716 adapted to Kubernetes.

**Value decisions**

Spill medium is the key difference between deployments: on SaaS (Cloud Run) the container filesystem is RAM-backed, so DuckDB spill and `/tmp` staging consume instance **memory** — that's why SaaS runs 24 GiB. On K8s the `/tmp` emptyDir is **disk-backed**, so spill costs node disk instead of RAM.

| Setting | cloud-native code default | SaaS (#25716) | This PR |
|---|---|---|---|
| `TRANSFER_DUCKDB_MEMORY_LIMIT` | `4GB` | `12GB` | `6GB` |
| `TRANSFER_DUCKDB_THREADS` | `2` | `4` | `4` |
| `TRANSFER_DUCKDB_TEMP_DIRECTORY` | unset (no spill) | `/tmp/duckdb-spill` | `/tmp/duckdb-spill` |
| `TRANSFER_DUCKDB_MAX_TEMP_DIRECTORY_SIZE` | unset (unbounded) | `4GB` | `15GB` |
| Node heap | Node default | pinned `4096` | `defaultNodeProcessMaxOldSpace: 4096` (single knob) |
| Container resources | n/a | 24 GiB / 6 vCPU | 12 GiB / 4 vCPU |
| `/tmp` bound (`tmpVolume.sizeLimit`) | n/a | n/a (RAM-bounded) | `""` (opt-in) |

Notes:

- **Chart today sets none of the env vars**, so the worker runs on the code defaults (4GB DuckDB, 2 threads, no spill directory) inside an 8 GiB / 2 vCPU container whose heap (50% rule → 4096 MB) already claims half of it — that's the gap this PR closes.
- **Memory budget (12 GiB)**: 4 GiB heap + 6GB DuckDB cap + ~2 GiB residual for native memory outside both (ADBC Go drivers, Arrow buffers, Node off-heap). The residual is not a new cost — the old 50% heap rule implicitly left the other 4 GiB for the ogr2ogr subprocess and native buffers; it's now just explicit. DuckDB's cap is not a reservation: legacy GDAL jobs run with the pool empty, so they keep at least the native room they have today.
- **Spill cap 15GB vs SaaS 4GB**: bounds only DuckDB's own spill → an oversized job fails with a classified error. SaaS keeps it small because it pays spill in RAM; ours is disk.
- **Heap fixed, not %**: heap demand comes from the GDAL pipeline, not container size — 50% of 12 GiB would hand the new memory to V8 instead of DuckDB. The percentage-derived path is removed for this component and `defaultNodeProcessMaxOldSpace` (bumped to 4096, today's effective heap) becomes the single heap knob: these keys are CARTO-managed (not exposed via KOTS or customization examples), and this release already changes the worker's resource model.
- **Threads/CPU**: matched pair (supersedes #874).
- **`/tmp` unbounded by default**: shared with legacy GDAL processing and BQ staging; a default bound could evict large legacy imports that complete today. Opt in (paired with `resources` ephemeral-storage) for node-disk protection.

All tunable via `values.yaml` / customizations. Numbers are starting points pending benchmarking on a test environment (same status as the SaaS values).

**Benefits**

Properly sized DuckDB memory budget with disk spill; oversized jobs fail cleanly instead of OOM-killing the worker. Legacy GDAL and `/tmp` behavior unchanged by default.

**Possible drawbacks**

Default footprint grows (memory 8→12 GiB, CPU 2→4). Tight deployments can lower the values keys.

**Additional information**

Validated with `helm lint` and `helm template`: default render diff vs `main` is only the intended lines (`NODE_OPTIONS` stays exactly `--max-old-space-size=4096`; `/tmp` stays `emptyDir: {}`); overrides and the heap-pin fallback verified. README regenerated (CI-enforced).

Shortcut: [sc-560525](https://app.shortcut.com/cartoteam/story/560525)


chart/values.yaml
```diff
@@ -1285,12 +1285,35 @@ importWorker:
   ## @param importWorker.customStartupProbe Custom startupProbe that overrides the default one
   ##
   customStartupProbe: {}
+  ## @param importWorker.nodeProcessMaxOldSpace Absolute memory in MB for node process max-old-space-size. Takes precedence over nodeProcessMaxOldSpacePercentage.
+  ## Pinned on purpose: DuckDB allocates native memory outside the V8 heap, so the heap must not scale with the container.
+  ## If unset (0), the percentage below applies — beware 50% of a DuckDB-sized container starves DuckDB's budget.
+  ##
+  nodeProcessMaxOldSpace: 4096
```
moicalcob: We also have `defaultNodeProcessMaxOldSpace` can you check why? 🙏🏼

chart/values.yaml
```diff
@@ -1300,11 +1323,11 @@ importWorker:
...
     limits:
-      memory: "8192Mi"
-      cpu: "2000m"
+      memory: "12288Mi"
+      cpu: "4000m"
     requests:
-      memory: "3072Mi"
-      cpu: "350m"
+      memory: "6144Mi"
+      cpu: "1000m"
```
moicalcob, line 1318: Remember to update:
- Spreadsheet that we have with the resources internally
- Public docs
- Create a ticket for release notes to notify customers ⚠️

chart/templates/import-worker/deployment.yaml
```diff
@@ -231,7 +231,13 @@ spec:
         {{- end }}
       volumes:
         - name: tmp
+          # Disk-backed on purpose: DuckDB spills here to relieve pod memory.
+          {{- if .Values.importWorker.tmpVolume.sizeLimit }}
+          emptyDir:
+            sizeLimit: {{ .Values.importWorker.tmpVolume.sizeLimit | quote }}
```
moicalcob: Why have we decided to set a size limit? have we checked in a dedicated k8s env how this limit is handled? If we import multiple files what are we storing in there and how we approach the limit?

I think that we need to clarify that a bit before adding limits as it could cause unknown issues no?

moicalcob (changes requested):
Just two blocking comments before we ship 🚢 😄

chart/templates/import-worker/deployment.yaml
mateo-di: I put this one as disabled by default but set the possibility of adding a limit to match SaaS behavior if necessary (since this is going to memory in SaaS there's a hard limit by default). I think we can remove it and add it in the future if necessary? I added it to have it into consideration 🤔

chart/values.yaml
mateo-di: Done: 

- https://app.shortcut.com/cartoteam/story/561596/release-notes-import-worker-resource-sizing-change-duckdb-engine-12gb-4cpu-disk-spill
- https://github.com/CartoDB/gitbook-documentation/pull/295

chart/values.yaml
mateo-di: Good catch. The first iteration of this PR played it safe: it added a new pinned knob and kept `nodeProcessMaxOldSpacePercentage` + `defaultNodeProcessMaxOldSpace` as fallbacks, assuming someone might depend on the old keys. From your comment i checked who actually consumes them: nobody.  

- They're CARTO-managed, not exposed in KOTS nor in any customization example.
So i consolidated the attributes instead of stacking knobs: the percentage path is gone and defaultNodeProcessMaxOldSpace is now the single heap setting, bumped to 4096 (today's effective heap, so the legacy GDAL path is unchanged).

- Why fixed instead of a % of the container: DuckDB allocates native memory outside the V8 heap, so growing the container must grow DuckDB's budget, not V8's share. Same conclusion as the SaaS module, which caps the heap at min(4096, 50%); we use the flat value and the values comment notes that sizing the worker below 8Gi means lowering this (and duckdb.memoryLimit) by hand.

chart/values.yaml
moicalcob: Let's please also ensure that product is aware of it for the release notes!

chart/templates/import-worker/deployment.yaml
moicalcob: I think that we can remove it as it doesn't work in the same way in SaaS and SH no?

chart/values.yaml
mateo-di: Will do!

chart/templates/import-worker/deployment.yaml
mateo-di: Yes for sure. Removed!

moicalcob (approved):
Great job 🚀

PR #894, opened 2026-07-13
feat(chart): fleet-wide trustedCACerts (custom CA) + soft-deprecate externalProxy.sslCA + KOTS field [sc-555247][sc-555248][sc-555249]

ManuelLR:
## Summary

Lets Self-Hosted trust a custom CA fleet-wide without a proxy, so a private-CA S3-compatible endpoint (MinIO/Ceph/OVH/OCI…) is trusted. Self-Hosted already wired `NODE_EXTRA_CA_CERTS` across the Node fleet, but the whole mechanism was gated on `externalProxy.sslCA` — semantically "the proxy's certificate". A customer with a private-CA endpoint but no proxy had no clean way to add it.

Delivers the **custom-CA work-stream** of the S3-compatible epic ([sc-544278](https://app.shortcut.com/cartoteam/epic/544278)). **Node plane only** — the native import engines (GDAL/DuckDB) don't honour `NODE_EXTRA_CA_CERTS` and are handled separately in [sc-561946](https://app.shortcut.com/cartoteam/story/561946).

## What changed

**New value `trustedCACerts`** (object, follows the chart's `appSecrets.*` `value` + `existing<Kind>` convention):

```yaml
trustedCACerts:
  value: ""                 # inline PEM(s), may be concatenated
  existingConfigmap:
    name: ""                # or reference a pre-existing ConfigMap
    key: ""                 # default ca.crt
```

- Feeds the same CA-bundle mount that `NODE_EXTRA_CA_CERTS` points at, across the whole fleet — **decoupled from `externalProxy.enabled`** so a no-proxy private-CA endpoint is trusted.
- **Legacy `externalProxy.sslCA` / `sslCAConfigmap` stay working**, gated on `externalProxy.enabled` as before (soft-deprecated, pointing at `trustedCACerts`). A breaking-change ticket for their removal in a future Helm major is tracked separately.
- Inline (`value`) and `existingConfigmap` are mutually exclusive (inline wins) — `NODE_EXTRA_CA_CERTS` is a single file.
- **Preflight (`tenant-requirements-checker`)** mounts the bundle too, so it trusts the private CA during bucket preflight.
- **KOTS Admin Console**: a new `trustedCaCertsValue` textarea (Advanced Config) → `trustedCACerts.value`, so the KOTS deployment methods can configure the CA too (the chart's `existingConfigmap` variant stays available for Helm users, just not surfaced in the Admin Console).
- **Resources renamed** to `trusted-ca-certs` (ConfigMap `<release>-trusted-ca-certs`, mount `/usr/src/certs/trusted-ca-certs`, volume `trusted-ca-certs`, helpers `carto.trustedCACerts.*`). The real HTTP-proxy plumbing (`computedConnectionString`, host/port/`NO_PROXY`) is untouched.

## Story links

- [sc-555247](https://app.shortcut.com/cartoteam/story/555247) — global `trustedCACerts` value feeding `NODE_EXTRA_CA_CERTS` (+ KOTS field)
- [sc-555248](https://app.shortcut.com/cartoteam/story/555248) — soft-deprecate `externalProxy.sslCA` / `sslCAConfigmap`
- [sc-555249](https://app.shortcut.com/cartoteam/story/555249) — close the preflight CA-mount gap

Native import engines (GDAL/DuckDB) trusting the private CA is **out of scope here** → [sc-561946](https://app.shortcut.com/cartoteam/story/561946).

## Validation

`helm template` across states: nothing set · legacy `sslCA` proxy-on/off · `trustedCACerts.value` · `existingConfigmap` · `sslCAConfigmap` proxy-on/off · both · inline-vs-CM precedence. In every state max one `NODE_EXTRA_CA_CERTS` per rendered doc; proxy-off legacy states render nothing (matching `main`). `helm lint` clean; README regenerated.

**Real TLS validation is deferred** to the provider-validation work-stream (SeaweedFS nightly) — not exercised here.

## Deployment impact

Additive. Existing AWS/GCS/Azure deployments and existing `externalProxy.sslCA` setups render as before (functionally equivalent; the CA resources were renamed, so not byte-identical, but the mounts/env point at the same content). No config change required on upgrade.

## Review focus

- The `trustedCACerts` shape (`value` + `existingConfigmap`) and the mutually-exclusive precedence.
- The restored `externalProxy.enabled` gating for the legacy sources (only the new sources are decoupled).
- KOTS field wiring in `manifests/kots-config.yaml` / `kots-helm.yaml`.


ManuelLR:
The 11 \`NODE_EXTRA_CA_CERTS\` "duplicate key" comments from gemini-code-assist are **false positives**. The two blocks were mutually exclusive: the added block was gated on `{{ not .Values.externalProxy.enabled }}` and the pre-existing one was nested **inside** the `{{ if .Values.externalProxy.enabled }}` block — so a ConfigMap never rendered the key twice (verified by `helm template`: max 1 occurrence per ConfigMap in every state).

That said, the two-block shape was needlessly confusing, so `65856fc` collapses them into a single `NODE_EXTRA_CA_CERTS` block gated only on `carto.customCA.enabled`, placed outside the proxy `if`. Behaviour is unchanged across all four states (nothing set / legacy `externalProxy.sslCA` / `customCA` / both) — only the key's position within `data` moved. Resolving these threads.

chart/templates/ai-api/configmap.yaml
```diff
@@ -50,11 +50,11 @@ data:
   NO_PROXY: {{ join "," .Values.externalProxy.excludedDomains | quote }}
   no_proxy: {{ join "," .Values.externalProxy.excludedDomains | quote }}
   {{- end }}
-  {{- if .Values.externalProxy.sslCA }}
-  NODE_EXTRA_CA_CERTS: {{ include "carto.proxy.configMapMountAbsolutePath" . | quote }}
-  {{- end }}
   NODE_TLS_REJECT_UNAUTHORIZED: {{ ternary "1" "0" .Values.externalProxy.sslRejectUnauthorized | quote }}
   {{- end }}
+  {{- if (include "carto.customCA.enabled" .) }}
```
ManuelLR: How we differentiative by name if the "customCA" is for outgoing connection or is the CA to be used in our component/http-router to receive traffic? I think the name is not clear enough.

Be careful, that renaming thing could change everywhere in that PR.

chart/templates/_commonChecks.tpl
```diff
@@ -141,9 +141,9 @@ Return common collectors for preflights and support-bundle
                 mountPath: {{ include "carto.redis.configMapMountDir" . }}
                 readOnly: false
               {{- end }}
-              {{- if and .Values.externalProxy.enabled .Values.externalProxy.sslCA }}
+              {{- if or .Values.customCA .Values.externalProxy.sslCA }}
```
ManuelLR:
```suggestion
              {{- if or .Values.customCA (and .Values.externalProxy.enabled .Values.externalProxy.sslCA) }}
```

right? I don't know why you decided to change the original condition.

chart/templates/_helpers.tpl
```diff
@@ -1331,10 +1331,41 @@ Return the proxy connection string if the config does not include the complete U
...
+Return "true" when there is any custom CA to trust fleet-wide, from any source:
+the global `customCA` value, the (deprecated) `externalProxy.sslCA`, or a
+pre-existing ConfigMap via `externalProxy.sslCAConfigmap.name`. This is the
+single source of truth for whether the CA bundle ConfigMap is created and the
+`proxy-ssl-ca` volume/mount + `NODE_EXTRA_CA_CERTS` env are rendered. It is
+intentionally decoupled from `externalProxy.enabled` so a private-CA endpoint
+(e.g. an S3-compatible store) is trusted even without a proxy.
+*/}}
+{{- define "carto.customCA.enabled" -}}
+{{- if or .Values.customCA .Values.externalProxy.sslCA .Values.externalProxy.sslCAConfigmap.name -}}
```
ManuelLR:
```suggestion
{{- if or .Values.customCA (and .Values.externalProxy.enabled ( or .Values.externalProxy.sslCA .Values.externalProxy.sslCAConfigmap.name)) -}}
```

I think you need to review all conditions changed because I think you lost some things.

chart/templates/_commonChecks.tpl
ManuelLR: Or why not use the templates already defined in `_helpers.tpl` ?

That applies to all the modifications in that file, not only that line.

chart/templates/import-worker/configmap.yaml
```diff
@@ -131,8 +131,8 @@ data:
   NO_PROXY: {{ join "," .Values.externalProxy.excludedDomains | quote }}
   no_proxy: {{ join "," .Values.externalProxy.excludedDomains | quote }}
   {{- end }}
-  {{- if or .Values.externalProxy.sslCA .Values.externalProxy.sslCAConfigmap.name }}
-  NODE_EXTRA_CA_CERTS: {{ include "carto.proxy.configMapMountAbsolutePath" . | quote }}
   {{- end }}
+  {{- if (include "carto.customCA.enabled" .) }}
+  NODE_EXTRA_CA_CERTS: {{ include "carto.customCA.configMapMountAbsolutePath" . | quote }}
```
ManuelLR: Are we sure that we don't need additional variables to work with DuckDB?

chart/templates/trusted-ca-certs-configmap.yaml
ManuelLR, line 1: I suppose that, at least, we can rename the files created for proxy, right?

chart/templates/NOTES.txt
```diff
@@ -15,6 +15,20 @@ It's no longer supported and you should use an external PostgreSQL instance inst
...
+{{- if .Values.externalProxy.sslCA }}
+
++-++-++-+ +-++-++-+ +-++-++-++-++-++-++-++-++-++-+
+|      externalProxy.sslCA IS DEPRECATED           |
++-++-++-+ +-++-++-+ +-++-++-++-++-++-++-++-++-++-+
+
+WARNING: `externalProxy.sslCA` is deprecated in favour of the global `customCA` value.
+It still works and is concatenated into the same trusted CA bundle, but it will be
+removed in a future release. Move your CA certificate(s) to the top-level `customCA`
+value, which trusts them fleet-wide via `NODE_EXTRA_CA_CERTS` regardless of whether an
```
ManuelLR: I think saying the variable name here don't add any kind of value.

chart/values.yaml
```diff
@@ -5595,6 +5595,14 @@ externalPostgresql:
   awsRdsRegion: ""
   aiProxyDatabaseName: "aiproxy"
 
+## @section Custom CA certificates
+## Extra CA certificate(s) to trust fleet-wide, mounted into every service via `NODE_EXTRA_CA_CERTS`.
+## Use this to trust a private/self-signed CA for any external endpoint the app talks to over TLS
+## (e.g. an S3-compatible object store, a custom proxy, etc.) without needing an external proxy.
+##
+## @param customCA Custom CA certificate(s) in PEM format to trust fleet-wide. May contain multiple concatenated PEM blocks. When set, a ConfigMap is mounted into every service and `NODE_EXTRA_CA_CERTS` points at it. Supersedes `externalProxy.sslCA`.
+customCA: ""
```
ManuelLR: Ok, that's good but we miss the option to directly obtain it from a confirmap as we have with the sslCAConfigmap right now. Is it really makes sense? is the sslCAConfigmap working today?

ManuelLR:
**Review addressed** in `06d5197` (on top of `65856fc` / `25a04b7`). Point by point:

1. **Naming — `customCA` was ambiguous** (outbound-trust vs a serving cert). Renamed to **`trustedCACerts`**, an object `{ value, existingConfigmap: { name, key } }` following the chart's `appSecrets.*` (`value` + `existing<Kind>`) convention. Propagated repo-wide — `customCA` / `carto.customCA` no longer appear anywhere.

2. **Lost `externalProxy.enabled` gate on the legacy sources** — restored. In `main` the CA mount/env lived inside `{{ if .Values.externalProxy.enabled }}`; the generalization had dropped that for `sslCA`/`sslCAConfigmap`. Now the legacy sources are gated on `externalProxy.enabled` again, and only the new `trustedCACerts.*` sources are decoupled from the proxy. Helper: `or trustedCACerts.value trustedCACerts.existingConfigmap.name (and externalProxy.enabled (or sslCA sslCAConfigmap.name))`.

3. **"Review all the changed conditions"** — audited every one. The proxy-off legacy states now render nothing (see states b2/e2 in the matrix below), matching `main`.

4. **"Use the helpers in `_commonChecks.tpl`"** — done; the four hand-written `or …` conditions there now use `carto.trustedCACerts.enabled` / `.hasInline`.

5. **DuckDB may need extra vars** — valid, same class as GDAL/`ogr2ogr` needing `GDAL_HTTP_CAINFO` (a native binary that ignores `NODE_EXTRA_CA_CERTS`). Investigating now whether DuckDB's httpfs respects the system trust store / `NODE_EXTRA_CA_CERTS` or needs its own CA pointer for TLS to the S3 endpoint. This PR is the Helm mechanism; the DuckDB piece will land as a follow-up. Leaving that thread open.

6. **Rename the proxy-named files/resources** — done now: `git mv externalproxy-configmap.yaml → trusted-ca-certs-configmap.yaml`; ConfigMap `<release>-trusted-ca-certs`, mount `/usr/src/certs/trusted-ca-certs`, volume `trusted-ca-certs`, helpers `carto.trustedCACerts.*`. The real HTTP-proxy plumbing (`computedConnectionString`, host/port/`NO_PROXY`) is untouched.

7. **NOTES.txt mentioning the internal var** — dropped the `NODE_EXTRA_CA_CERTS` reference; it describes the behaviour now, not the mechanism.

8. **Missing ConfigMap source (parity with `sslCAConfigmap`)** — verified `sslCAConfigmap` works today, so added `trustedCACerts.existingConfigmap.{name,key}`. Inline (`value`) and `existingConfigmap` are mutually exclusive, inline wins — `NODE_EXTRA_CA_CERTS` points at a single file.

Validated with `helm template` across 9 states (nothing · legacy `sslCA` proxy-on/off · `trustedCACerts.value` · `existingConfigmap` · `sslCAConfigmap` proxy-on/off · both · inline-vs-CM precedence); max one `NODE_EXTRA_CA_CERTS` per rendered doc; `helm lint` clean; README regenerated. Because the resources were renamed, the legacy states are functionally equivalent to `main` (same mounts/env) rather than byte-identical.

chart/templates/import-worker/configmap.yaml
ManuelLR: Addressed. Investigated the concern: the CA gap is common to all native engines (GDAL via libcurl, DuckDB via OpenSSL) — neither honours `NODE_EXTRA_CA_CERTS`, and it is a pre-existing gap (see `proxy-https.env`), not introduced here. GDAL is now wired in this PR (`GDAL_HTTP_CAINFO`, commit fab4c1a); DuckDB gets `SET ca_cert_file` in cloud-native #26305. Both tracked under sc-555242. This PR's `trustedCACerts` covers the Node plane; the native-engine env vars just point at the same mounted file. Note: `GDAL_HTTP_CAINFO`/`ca_cert_file` replace the system bundle (not append) — fine for air-gapped/private-endpoint targets; mixed public+private surfaces concatenate public roots into `trustedCACerts.value`.

chart/templates/import-api/configmap.yaml
```diff
@@ -131,8 +131,14 @@ data:
...
-  NODE_EXTRA_CA_CERTS: {{ include "carto.proxy.configMapMountAbsolutePath" . | quote }}
   {{- end }}
+  {{- if (include "carto.trustedCACerts.enabled" .) }}
+  NODE_EXTRA_CA_CERTS: {{ include "carto.trustedCACerts.configMapMountAbsolutePath" . | quote }}
+  ## GDAL/ogr2ogr reads directly from the (S3) endpoint via /vsicurl/ and uses libcurl, not Node's
+  ## TLS stack, so point its CA bundle at the same mounted private CA. NOTE: GDAL_HTTP_CAINFO
+  ## REPLACES the system bundle (it does not append), so the trusted-ca-certs file becomes the ONLY
+  ## CA libcurl trusts for imports. This is intended for private-CA endpoints (the air-gapped case);
+  ## see the `trustedCACerts` section in values.yaml.
+  GDAL_HTTP_CAINFO: {{ include "carto.trustedCACerts.configMapMountAbsolutePath" . | quote }}
```
ManuelLR: First, remove all that comment. Then, you never said that it's going to replace the system certs and that it's a problem. Do you know if it support more than one file in that variable? If not, please, don't define it.

chart/templates/import-worker/configmap.yaml
```diff
@@ -131,8 +131,14 @@ data:
...
-  NODE_EXTRA_CA_CERTS: {{ include "carto.proxy.configMapMountAbsolutePath" . | quote }}
   {{- end }}
+  {{- if (include "carto.trustedCACerts.enabled" .) }}
+  NODE_EXTRA_CA_CERTS: {{ include "carto.trustedCACerts.configMapMountAbsolutePath" . | quote }}
+  ## GDAL/ogr2ogr reads directly from the (S3) endpoint via /vsicurl/ and uses libcurl, not Node's
+  ## TLS stack, so point its CA bundle at the same mounted private CA. NOTE: GDAL_HTTP_CAINFO
+  ## REPLACES the system bundle (it does not append), so the trusted-ca-certs file becomes the ONLY
+  ## CA libcurl trusts for imports. This is intended for private-CA endpoints (the air-gapped case);
+  ## see the `trustedCACerts` section in values.yaml.
+  GDAL_HTTP_CAINFO: {{ include "carto.trustedCACerts.configMapMountAbsolutePath" . | quote }}
```
ManuelLR: Same as https://github.com/CartoDB/carto-selfhosted-helm/pull/894/changes#r3576426840

chart/templates/_commonChecks.tpl
```diff
@@ -542,6 +541,10 @@ Return customer values to use in preflights and support-bundle
   - name: WORKSPACE_IMPORTS_STORAGE_ACCOUNT
     value: {{ .Values.appConfigValues.azureStorageAccount | quote }}
   {{- end }}
+  {{- if and (not .Values.externalProxy.enabled) (include "carto.trustedCACerts.enabled" .) }}
```
ManuelLR: Why that condition is different of others? I think that this is already defined in this file so it's not required, right?

chart/templates/import-api/configmap.yaml
```diff
@@ -131,8 +131,14 @@ data:
...
-  NODE_EXTRA_CA_CERTS: {{ include "carto.proxy.configMapMountAbsolutePath" . | quote }}
   {{- end }}
+  {{- if (include "carto.trustedCACerts.enabled" .) }}
+  NODE_EXTRA_CA_CERTS: {{ include "carto.trustedCACerts.configMapMountAbsolutePath" . | quote }}
+  ## GDAL/ogr2ogr reads endpoints directly over libcurl (e.g. an S3-compatible store via /vsicurl/),
+  ## not Node's TLS stack. GDAL_HTTP_CAPATH maps to CURLOPT_CAPATH, which OpenSSL searches IN ADDITION
+  ## TO the system bundle (already pointed at by CURL_CA_BUNDLE baked into the image), so both public-
+  ## and private-CA endpoints stay trusted. The directory holds the hash-named symlinks produced by
+  ## the `trusted-ca-rehash` initContainer, which rehashes the private CA into a shared emptyDir.
+  GDAL_HTTP_CAPATH: "/usr/src/certs/trusted-ca-capath"
```
ManuelLR: But... that path should point to the dynamic value as the NODE_EXTRA_CA_CERTS variable, right?

chart/templates/import-api/deployment.yaml
```diff
@@ -69,6 +69,34 @@ spec:
...
+          {{- if .Values.importApi.resources }}
+          resources: {{- toYaml .Values.importApi.resources | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - name: trusted-ca-certs
+              mountPath: {{ include "carto.trustedCACerts.configMapMountDir" . }}
+              readOnly: true
+            - name: trusted-ca-capath
+              mountPath: /tmp/trusted-ca-capath
+        {{- end }}
```
ManuelLR: Drop that, (also in import-worker), we never approve that approach

chart/templates/import-api/deployment.yaml
```diff
@@ -201,9 +229,12 @@ spec:
...
             {{- end }}
-            {{- if and .Values.externalProxy.enabled (or .Values.externalProxy.sslCA .Values.externalProxy.sslCAConfigmap.name) }}
-            - name: proxy-ssl-ca
-              mountPath: {{ include "carto.proxy.configMapMountDir" . }}
+            {{- if (include "carto.trustedCACerts.enabled" .) }}
+            - name: trusted-ca-certs
+              mountPath: {{ include "carto.trustedCACerts.configMapMountDir" . }}
+              readOnly: true
+            - name: trusted-ca-capath
+              mountPath: /usr/src/certs/trusted-ca-capath
```
ManuelLR: I suppose you should also drop this.

chart/templates/import-api/deployment.yaml
```diff
@@ -243,10 +274,12 @@ spec:
...
         {{- end }}
-        {{- if and .Values.externalProxy.enabled (or .Values.externalProxy.sslCA .Values.externalProxy.sslCAConfigmap.name) }}
-        - name: proxy-ssl-ca
+        {{- if (include "carto.trustedCACerts.enabled" .) }}
+        - name: trusted-ca-certs
           configMap:
-            name: {{ include "carto.proxy.configMapName" . }}
+            name: {{ include "carto.trustedCACerts.configMapName" . }}
+        - name: trusted-ca-capath
+          emptyDir: {}
```
ManuelLR: And this

manifests/kots-config.yaml
```diff
@@ -739,6 +739,18 @@ spec:
...
+          title: Trusted CA certificate(s)
+          type: textarea
+          help_text: |-
+            PEM-encoded CA certificate(s) to trust for outbound TLS connections, i.e. verifying the servers CARTO connects to when they are signed by a private or self-signed CA (for example a private-CA S3-compatible object store). You may paste multiple concatenated PEM blocks. Leave empty if not needed.
+        - name: trustedCaCertsExistingConfigmapName
+          title: Trusted CA ConfigMap name
+          type: text
+          help_text: |-
+            Alternatively, reference a pre-existing ConfigMap in the release namespace holding the CA bundle (key `ca.crt`). Ignored when a certificate is pasted above.
+
```
ManuelLR: That options isn't required at all so please, drop it (and the kots-helm.yaml block related with that)

manifests/kots-config.yaml
```diff
@@ -739,6 +739,18 @@ spec:
           value: 'repl{{ PrivateCACert }}'
           readonly: true
 
+        ## Trusted CA certificates
+        - name: trustedCaCertsValue
+          title: Trusted CA certificate(s)
+          type: textarea
+          help_text: |-
+            PEM-encoded CA certificate(s) to trust for outbound TLS connections, i.e. verifying the servers CARTO connects to when they are signed by a private or self-signed CA (for example a private-CA S3-compatible object store). You may paste multiple concatenated PEM blocks. Leave empty if not needed.
```
ManuelLR: I think we should indicate that this is going to append to the standard certificate supported list (or whatever they are called). That syntax could cause confusion and be understanded as "replace the standard supported ones" instead of the "append" or "extra" current method

mateo-di (commented):
Manolo, this LGTM but I remember seeing a thread (couldn't find it) where Alex was saying that we didn't need to implement this unless we find an actual case where a customer need it 🤔 

Apart from that question, did you test this in a K8s ded? I couldn't find that you used the `release-changes` label

PD: found the thread: https://cartoteam.slack.com/archives/C0B5YSDCFC6/p1784031711062479?thread_ts=1784017482.118149&cid=C0B5YSDCFC6

ManuelLR:
### Validated on a dedicated-k8s ✅

Deployed this branch's chart on a real dedicated-k8s (`ded-k8s-07`) by pointing its Replicated customer at this branch's channel (`feature_sc_555247_helm_custom_ca_global_value`) via the `override-channel` mechanism — cloud-native PR CartoDB/cloud-native#26544 (testing-only, not to be merged).

**Result — clean:**
- `Reserve env` (install + **preflight** / `tenant-requirements-checker`): ✅ success
- `CI - Backend` (branch images) + `CI - Frontend`: ✅ success

So the chart changes here (fleet-wide `trustedCACerts` + `NODE_EXTRA_CA_CERTS`, soft-deprecated `externalProxy.sslCA`, preflight CA mount, resources renamed to `trusted-ca-certs`) install and pass preflight on a real dedicated with **no regression on a standard install** (where `trustedCACerts` is unset — the additive/backward-compatible path).

_Scope note:_ a standard dedicated-k8s doesn't point storage at an S3-compatible endpoint with a private CA, so `trustedCACerts` stays empty and the trust path isn't exercised end-to-end here. Real TLS-against-a-private-CA validation stays with the provider-validation work-stream (SeaweedFS nightly).

ManuelLR:
@mateo-di thanks! Two things:

**On Álex's "if and when":** that thread is specifically about the **native import engines (GDAL + DuckDB)** trusting a private CA — i.e. [sc-561946](https://app.shortcut.com/cartoteam/story/561946), which is exactly what got pulled out of this PR and parked (it needs read-only-FS gymnastics — initContainer, cert rehash/combined bundle — that we didn't want to approve). Álex's "if and when" applies to *that*.

**This PR is a narrower, different thing — the Node plane only.** It doesn't add a new capability so much as *generalize an existing one*: `NODE_EXTRA_CA_CERTS` was already wired fleet-wide, but gated behind `externalProxy.sslCA` ("the proxy's cert"). This just decouples it from the proxy so a **no-proxy private-CA S3 endpoint** is trusted — which the S3-compatible epic's own **preflight / HeadBucket / `/storage/sign` / thumbnails** need to work against such an endpoint. It's additive, backward-compatible, and byte-equivalent when unset, so I don't think it falls under that deferral.

**On the K8s-ded test:** done — validated on `ded-k8s-07` (details in the comment above). Install + preflight + deploy all clean, no regression on a standard install.

mateo-di (approved):
@ManuelLR approving this one! Lets make sure that we are aligned with product!

PR #901, opened 2026-07-20
feat(ci): fail lint when chart limits exceed the admission ceiling [sc-562980]

mateo-di:
## Summary
Fail the `Lint codebase` run at PR time when any chart component's `resources.limits` exceed the container-limit ceiling admitted on CARTO-managed clusters, so an oversized limit is caught in review instead of silently wedging a Deployment at 0 replicas after ship.
Story: [sc-562980]

## What Changed
**Added:**
- `.github/scripts/check-resource-limits.sh` — `yq` extracts every `resources.limits`, `awk` normalizes CPU (`m`/bare cores) and memory (`Mi`/`Gi`), exits 1 listing any component over the ceiling. The ceiling is passed in via `MAX_CPU_M` / `MAX_MEMORY_MI` (currently `4000m` / `12288Mi`).
- `.github/scripts/check-resource-limits-test.sh` + `.github/scripts/testdata/resource-limits-fixture.yaml` — a self-test that proves the check still catches offenders.
- Two steps in the existing PR-time `check-helm-resources-changed` job (`.github/workflows/lint-codebase.yaml`): run the self-test, then run the check against `chart/values.yaml`.

**Modified:**
- `CLAUDE.md` — documents the ceiling, why exceeding it wedges a Deployment, and how to raise it.

**Removed:**
- N/A vs `main`. An earlier revision was notify-only (a Python + PyYAML script plus PR-comment and Slack wiring); it was replaced — see Architectural Context.

## Architectural Context
EAD: N/A - small change

**Architectural Decisions Made:**
- **Fail vs notify:** a failing check *is* both the guardrail and the notification. That removes the whole notify-only pipeline (script, GitHub outputs, report file, PR comment, conditional Slack text) the first revision needed.
- **Where it lives:** appended to the existing PR-time lint job rather than the push-to-`main` resources workflow. Wiring it into that workflow would also have made its `@here` Slack notice fire on every PR push that touched a resource value.
- **Script, not inline:** the check started as an inline `yq | awk` block. Nothing proved it worked, so a later "simplification" of the unit normalization could have turned the guard into a silent no-op. Extracting it to a shell script makes it testable while keeping the original constraint that motivated dropping the Python version — `yq` and `awk` are both preinstalled, so there is still no `pip install`.
- **Known trade-off:** the ceiling is a constant mirrored from the cluster policy and can drift. Reading it live isn't simpler (private source, token). A hard-failing check makes any drift visible at the one moment it matters — when someone raises a limit past it.

## Review Focus Areas
**Critical areas:**
1. `.github/scripts/check-resource-limits.sh` — the `awk` unit normalization, and that `set -euo pipefail` makes a `yq` failure fail the step rather than pass silently.
2. `.github/scripts/check-resource-limits-test.sh` — whether the fixture covers the shapes that matter.

**Safe to skip**: `CLAUDE.md` prose.

## Deployment Impact
- [ ] Pure Helm installs
- [ ] Replicated / KOTS installs (Admin Console, embedded cluster)
- [ ] Both install paths
- [x] Not applicable (docs, CI only) — CI-only, no runtime change.

## Migration & Breaking Changes
- [x] No migrations or breaking changes

## Security Considerations
- [x] No security impact

## Performance Impact
- [x] No performance impact

## Tests
The self-test runs in CI ahead of the check itself, so the guard is verified before it is trusted.

Fixture coverage — each entry is one shape the check has to get right:

| Shape | Expected |
|---|---|
| `requests` only, no `limits` | not flagged |
| `limits: {cpu: null, memory: null}` | not flagged, no crash |
| `limits: {}` | not flagged, no crash |
| exactly at the ceiling (`4000m`/`12288Mi`) | not flagged — `importWorker` really sits here |
| bare cores under the ceiling (`cpu: 2`) | not flagged |
| milliCPU over (`8000m`) | flagged |
| bare cores over (`cpu: 6`) | flagged |
| memory over in `Gi` (`16Gi`) | flagged |
| memory over in `Mi` (`20480Mi`) | flagged |
| over-limit entry nested in a list (`sidecars[0]`) | flagged |

Mutation-tested the self-test against three plausible regressions — it fails on each:

| Mutation | Caught by |
|---|---|
| drop the `Gi` → `Mi` conversion | `expected offender 'overMemoryGi' was not reported` |
| drop the bare-cores `*1000` | `expected offender 'overCpuBareCores' was not reported` |
| change `>` to `>=` | `'atCeiling' must not be flagged but was` |

- [ ] Renders on both install paths — N/A, no chart change
- [ ] Validated on a real install — new install — N/A, CI-only
- [ ] Validated on a real install — upgrade from the released chart — N/A, CI-only
- [x] Chart tests / preflight specs added or updated — CI self-test added
- [x] Edge cases verified: see the fixture table above
- [ ] No tests needed

Confirmed green on the runner (not just locally), using the runner's preinstalled `yq`:
```
Self-test the resource-limit check       -> check-resource-limits.sh self-test: all assertions passed.
Check resource limits ... ceiling        -> All resources.limits in chart/values.yaml are within 4000m CPU / 12288Mi.
```

## Dependencies
- None

## How to Validate
1. `./.github/scripts/check-resource-limits-test.sh` — passes.
2. `./.github/scripts/check-resource-limits.sh chart/values.yaml` — passes.
3. Raise any component's `resources.limits.cpu` in `chart/values.yaml` above `4000m` and re-run step 2 — it exits 1 and names the component.

## Screenshots/Demos
N/A — CI output quoted above.

## AI-Generated Code Notice
- [x] This PR contains AI-generated code
- [x] Areas requiring extra verification: the `awk` unit normalization; the fixture's claim to cover the real shapes in `chart/values.yaml`
- [ ] Not applicable

## Coding Standards Compliance
- [x] Changes follow the repo conventions
- [ ] Chart documentation updated (via helm-readme-generator) — N/A, no `values.yaml` change
- [ ] Chart templates linted and validated — N/A, no template change
- [ ] Customer-set values wired end to end — N/A
- [ ] New values carry a `## @param` comment — N/A
- [x] Version fields left to the release bot
- [x] No secrets, internal hostnames/project IDs, or customer data added (public repo)

## Checklist
- [x] PR title follows convention
- [x] Shortcut story linked
- [x] One issue per PR
- [x] Appropriate labels applied
- [x] Reviewers assigned (or auto-assigned)
- [x] AI review findings addressed (if applicable)


mateo-di:
### Ceiling constant verified against the live admission policy

The `MAX_CPU_M` / `MAX_MEMORY_MI` values in this PR are mirrored from the Gatekeeper constraint, and the mirror is the known trade-off called out in Architectural Context. Checked them against a live CARTO-managed cluster running `container-must-have-limits`:

| | live constraint `parameters` | this PR |
|---|---|---|
| CPU | `cpu: "4"` | `MAX_CPU_M: 4000` |
| Memory | `memory: 12Gi` | `MAX_MEMORY_MI: 12288` |

Both match, so the check is calibrated to what admission actually enforces today — and `importWorker` sitting exactly at `4000m` / `12288Mi` is right at the line, which is why the comparison is `>` and not `>=` (the fixture asserts that).

### Review-thread status

The three Gemini comments on this PR were filed against `.github/scripts/check-gatekeeper-limits.py`, from the first revision. That script no longer exists — the implementation was replaced with `yq` + `awk` — and GitHub has already marked all three threads outdated and resolved. Worth noting they are also **non-issues** in the current design rather than merely stale:

| Gemini concern | Status in the `yq`/`awk` implementation |
|---|---|
| `float('None')` raises `ValueError` on null/empty limits and crashes the script | Does not apply. `awk` coerces the non-numeric string to `0`, which is below any ceiling, so unset limits are skipped. Asserted by the `nullLimits` / `emptyLimits` / `requestsOnly` fixture entries. |
| `mem_to_mi` crashes the same way | Same — asserted by the same fixture entries. |
| `walk` only traverses dicts, so `sidecars` / `initContainers` lists are ignored | Does not apply. `yq`'s `..` recursive descent traverses lists natively. Asserted by the `nestedList.sidecars[0]` fixture entry, which is reported as `nestedList.sidecars.0`. |

The underlying point was fair, though — nothing *proved* any of that. That gap is what the new self-test closes, and it is mutation-tested against three plausible regressions (details in the PR body).

PR #846, opened 2026-04-06
fix(import-api,import-worker): wire export bucket envs for pdf export [sc-540843]

mateo-di:
## Summary
Wire the selfhosted export bucket settings into `import-api` and `import-worker` so PDF export can resolve the configured export bucket in dedicated-k8s deployments. Mirror the existing `maps-api` and `sql-worker` export wiring, including S3 secret-backed credentials and the optional GCS credentials file path, while keeping custom branding bucket wiring opt-in.

Story: [sc-540843](https://app.shortcut.com/cartoteam/story/540843)

## What Changed
**Added:**
- Export bucket environment variables in `import-api` and `import-worker` configmaps
- S3 export access key secret wiring for `import-api` and `import-worker`
- Optional `EXPORTS_GCS_CREDENTIALS_KEY_FILENAME` wiring for both import services

**Modified:**
- `chart/templates/import-api/configmap.yaml`
- `chart/templates/import-api/deployment.yaml`
- `chart/templates/import-api/secret.yaml`
- `chart/templates/import-worker/configmap.yaml`
- `chart/templates/import-worker/deployment.yaml`
- `chart/templates/import-worker/secret.yaml`
- Normalized the import service secret-name lists to the same alphabetical order across `deployment.yaml` and `secret.yaml`

**Removed:**
- Default `WORKSPACE_BRANDINGS_BUCKET` mapping to the thumbnails bucket from the import services

## Architectural Context
EAD: N/A - small change

**Architectural Decisions Made:**
- Keep the source of truth for export bucket values unchanged and only mirror the already-existing export env pattern into the runtime services used by PDF export
- Match `import-api` and `import-worker` export wiring to the existing `maps-api` and `sql-worker` behavior to avoid introducing a separate selfhosted export path
- Leave custom branding bucket configuration opt-in so selfhosted matches the latest upstream `cloud-native` behavior

## Review Focus Areas
**Critical areas** (require thorough review):
1. `chart/templates/import-api/*` - export env parity with the PDF export runtime path
2. `chart/templates/import-worker/*` - export env parity for the async subscriber path
3. Secret-backed S3 export credential wiring - confirm it matches the existing `maps-api`/`sql-worker` pattern
4. Secret list ordering - confirm the import deployment and secret templates stay aligned for maintainability

**Safe to skip**: unrelated chart templates and any branding bucket defaults, since this PR intentionally does not configure them by default

## Deployment Impact
- [ ] SaaS only
- [x] Selfhosted only
[...]

PR #903, opened 2026-07-21
fix(chart): wire EXPORTS_GCS_BUCKET_NAME into import-api and import-worker

mateo-di:
## Summary

`import-api` and `import-worker` were the only backend services whose configmaps did not render `EXPORTS_GCS_BUCKET_NAME` (`workspace-api`, `maps-api` and `sql-worker` already do). The DuckDB export engine that backs the interactive data-export flow (`import-api` `POST /v3/exports`, processed by `import-worker`) stages BigQuery `EXPORT DATA` output to the bucket named by `EXPORTS_GCS_BUCKET_NAME`, and falls back to the imports bucket when that variable is unset. On deployments that configure **separate** import and export buckets (i.e. "configure your own buckets"), the export was therefore staged in the imports bucket, which does not carry the export bucket's IAM — so BigQuery exports failed with a permission-denied writing to GCS.

This wiring existed briefly before and was removed as unused when the PDF-export path stopped reading it (#846); the DuckDB export engine reintroduced the read afterward, so it needs to be restored.

## Story

[sc-563200](https://app.shortcut.com/cartoteam/story/563200)

## What changed

Render `EXPORTS_GCS_BUCKET_NAME` from `appConfigValues.workspaceExportsBucket` in:

- `chart/templates/import-api/configmap.yaml`
- `chart/templates/import-worker/configmap.yaml`

Mirrors the existing wiring in `sql-worker`, `workspace-api` and `maps-api`. One line each; no default value change (when `workspaceExportsBucket` is unset it renders empty, preserving the current fallback behaviour).

## Deployment impact

Selfhosted only. On a deployment where import and export buckets are the same, behaviour is unchanged. On a deployment with separate buckets, the interactive export now stages in the configured export bucket instead of the imports bucket. No migration.

## Validation

`helm template` with `workspaceExportsBucket` set renders the variable in both configmaps:

```
# import-api & import-worker configmaps
EXPORTS_GCS_BUCKET_NAME: "…-private-export-bucket"
IMPORT_DATA_IMPORTS_BUCKET: "…-private-client-bucket"
```

## Review focus

Confirm `EXPORTS_GCS_BUCKET_NAME` should map to `workspaceExportsBucket` for the import services (same source the other three services use).

### DIFF
--- a/chart/templates/import-api/configmap.yaml
+++ b/chart/templates/import-api/configmap.yaml
@@ -82,6 +82,7 @@ data:
   CARTO_SELFHOSTED_AWS_RDS_METADATA_REGION: {{ .Values.externalPostgresql.awsRdsRegion | quote }}
   CARTO_SELFHOSTED_AWS_EKS_POD_IDENTITY_METADATA_DB_ENABLED: {{ .Values.externalPostgresql.awsEksPodIdentityEnabled | quote }}
   {{- end }}
+  EXPORTS_GCS_BUCKET_NAME: {{ .Values.appConfigValues.workspaceExportsBucket | quote }}
   IMPORT_AWS_CUSTOM_BUCKET_ROLE_ARN: {{ .Values.appConfigValues.importAwsRoleArn | quote }}
   IMPORT_DATA_IMPORTS_BUCKET: {{ .Values.appConfigValues.workspaceImportsBucket | quote }}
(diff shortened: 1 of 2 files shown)


waldnzwrld (approved):
Good catch

ManuelLR (approved):

PR #909, opened 2026-07-28
fix(chart): flatten default image repositories for router, router-metrics and cdn-invalidator-sub [sc-422038]

waldnzwrld:
Relies on https://github.com/CartoDB/cloud-native/pull/26795 to be shipped. 

**Description of the change**

Flattens the default `image.repository` for the four components that still
pointed at a nested path in the registry, so they match the flat layout every
other component already uses:

| Value | Before | After |
| --- | --- | --- |
| `router.image.repository` | `router/router-http` | `router-http` |
| `routerMetrics.image.repository` | `router/router-metrics` | `router-metrics` |
| `upgradeCheck.image.repository` | `router/router-http` | `router-http` |
| `cdnInvalidatorSub.image.repository` | `consumers/cdn-invalidator-sub` | `cdn-invalidator-sub` |

`chart/README.md` is regenerated to match. No template, helper or KOTS change is
needed — the KOTS layer does not expose image repositories, so both install
paths (plain Helm and Replicated/KOTS) pick this up from `chart/values.yaml`
alone.

Note the remaining `consumers/cdn-invalidator-sub` occurrences in
`chart/values.yaml` and `chart/templates/cdn-invalidator-sub/deployment.yaml`
are the *in-container* `dist/` path in the container args, not an image
repository, and are intentionally left untouched.

**Benefits**

- Chart defaults resolve to images that actually exist at the flattened
  registry paths, so a default install pulls successfully.
- All image repositories now follow one convention, which makes the values
  easier to read and to override.

**Possible drawbacks**

- This changes chart defaults: an install that relied on the previous nested
  default now pulls from a different path. Anyone mirroring images into a
  private registry needs their mirror to expose the flattened names (or to set
  `image.repository` explicitly).
- Because `upgradeCheck` shares the router image, the upgrade-check hook picks
  up the same change.

**Applicable issues**

  - [sc-422038]

**Additional information**

Verified locally with `helm template` on both paths — plain and
`--set replicated.enabled=true` — confirming the rendered image references
resolve to the flattened repositories.


mateo-di:
hey @waldnzwrld do we still need this PR to be shipped?

waldnzwrld:
> hey @waldnzwrld do we still need this PR to be shipped?

@moicalcob had asked that I hold off on tis as it was part of a larger body of work. I think it is still viable, but I haven't made a push to ship. I can close it or we could ship it when ready.

PR #914, opened 2026-07-31
feat(chart): expose Workflows HTTP Request IP filtering config [sc-479237]

mateo-di:
## Summary
Exposes the Workflows HTTP Request component's IP filtering as chart config. The component filters outbound requests by IP range (private and reserved ranges are blocked by default), and until now Helm operators could only change that through the generic `extraEnvVars` escape hatch. This adds three first-class `appConfigValues` so it can be configured, and documented, like any other knob.

Story: [sc-479237](https://app.shortcut.com/cartoteam/story/479237)
Platform side (workspace-api): CartoDB/cloud-native#26880

## What changed
Three new `appConfigValues`, wired into the workspace-api ConfigMap:

| Value | Default | Env var | Effect |
|-------|---------|---------|--------|
| `workflowsHttpRequestRestrictIps` | `true` | `WORKSPACE_HTTP_REQUEST_RESTRICT_IPS` | Master switch for the filter |
| `workflowsHttpRequestAllowedIpRanges` | `[]` | `WORKSPACE_HTTP_REQUEST_ALLOWED_IP_RANGES` | Allowlist ranges to reach internal endpoints; checked ahead of the blocklist |
| `workflowsHttpRequestRestrictedIpRanges` | `[]` | `WORKSPACE_HTTP_REQUEST_RESTRICTED_IP_RANGES` | Override the default blocklist entirely |

Defaults preserve current behaviour: the filter is already on by default in the app, so an install that sets none of these renders `RESTRICT_IPS: "true"` and no range lists, exactly as today.

## Design notes
- The switch always renders as a real boolean, so it can never be emitted blank; the app rejects an empty value at boot. That is why it is not wrapped in an `{{- if }}`.
- The two range lists render only when non-empty (`join ","`), so an unset list falls back to the app's own defaults rather than freezing a list into the chart.

## Validation
- `helm template` under four value combinations (default / allowlist set / filter off / blocklist override): each renders the expected env vars and `RESTRICT_IPS` is never blank.
- `helm lint`: clean.
- README values table regenerated with the helm-readme-generator (matches the CI parity check).

## Follow-up (not in this PR)
- KOTS Admin Console fields for these values, for Replicated installs (mirrors how the S3-compatible config was split across #887 → #888).

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -59,6 +59,9 @@ To install, upgrade or uninstall this chart, please refer to [the root README.md
 | `appConfigValues.ldsIsolineProvider`                               | The default LDS provider for isolines                                                                                  | `""`                   |
 | `appConfigValues.logLevel`                                         | The log level used in CARTO application                                                                                | `info`                 |
 | `appConfigValues.deploymentType`                                   | The deployment type used in CARTO application                                                                          | `helm`                 |
+| `appConfigValues.workflowsHttpRequestRestrictIps`                  | Block the Workflows HTTP Request component from reaching private and reserved IP ranges. Enabled by default.           | `true`                 |
+| `appConfigValues.workflowsHttpRequestAllowedIpRanges`              | CIDR ranges the HTTP Request component may reach even when otherwise blocked. Checked before the blocklist.            | `[]`                   |
+| `appConfigValues.workflowsHttpRequestRestrictedIpRanges`           | Override the default IP blocklist for the HTTP Request component. When set, it replaces CARTO's defaults.              | `[]`                   |
 
 ### CARTO Replicated parameters
 
--- a/chart/templates/workspace-api/configmap.yaml
+++ b/chart/templates/workspace-api/configmap.yaml
@@ -64,6 +64,14 @@ data:
   USING_EXTERNAL_REDIS: "true"
   {{- end }}
   WORKSPACE_ERROR_RESPONSE_STACK_TRACE: {{ .Values.cartoConfigValues.enableErrorResponseStackTrace | quote }}
+  {{- /* Always render the switch as a real boolean: an empty value makes workspace-api throw at boot. */}}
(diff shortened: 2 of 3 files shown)


waldnzwrld (commented):
The new block has no nil guard.

  Lists — chart/templates/workspace-api/configmap.yaml:69 and :72. A customer writing the key with no value:
  appConfigValues:
    workflowsHttpRequestAllowedIpRanges:
  hard-fails templating: error calling len: len of nil pointer. Install never starts.

  The switch — :67. Same shape, workflowsHttpRequestRestrictIps: empty, renders:
  WORKSPACE_HTTP_REQUEST_RESTRICT_IPS:
  That parses as None, and ConfigMap.data is map[string]string, so the API server rejects the object. This directly contradicts
  the PR's design note — "always renders as a real boolean, so it can never be emitted blank." What protects it is the
  values.yaml default, not the template; | quote on nil yields nothing, not "". Since the note is the stated reason the switch
  isn't wrapped in an {{- if }}, the reasoning behind that choice doesn't hold as written.

  In fairness: gt (len …) 0 + join "," is copied verbatim from externalProxy.excludedDomains, so the PR follows house
  convention, and that pattern carries the identical latent bug — proxy enabled plus an empty excludedDomains: fails the same
  way. The difference that matters: excludedDomains sits behind {{- if .Values.externalProxy.enabled }}, so it's only reachable
  for customers who opt into the proxy. The new block has no outer guard, so it's on the default render path for every install.

  Fix is a truthiness test, which covers nil, empty list, and empty string at once:
  {{- if .Values.appConfigValues.workflowsHttpRequestAllowedIpRanges }}
  plus | default true | quote on the switch. Tradeoff: diverges from the gt (len …) idiom used in ~13 other places, so it reads
  as inconsistent unless you note why.

  Likelihood is medium-low — it needs a customer to leave a key valueless, which is plausible when hand-assembling values.yaml
  from README snippets, but isn't the common path. Blast radius is high: install-time failure on infrastructure you can't
  quickly redeploy.

PR #928, opened 2026-08-21
fix(selfhosted): keep GCS-only defaults and vars off non-GCS storage providers [sc-569088]

mateo-di:
**Description of the change**

Two small fixes closing one defect class: values meant for the GCS storage provider reaching deployments running on a different provider.

📎 Full write-up of the bucket wiring, the leak, and the follow-ups (internal — diagrams + live values): https://claude.ai/code/artifact/468689ba-ed4a-47f5-aa4a-984db4b589ae

1. `manifests/kots-config.yaml` — the `storageBucketsGcpProjectId` default now requires `storageBucketsBehaviour == custom_gcs` in addition to `replicatedLicenseType == dev`, matching the `storageBucketsWorkspaceImports` / `storageBucketsThumbnails` siblings. KOTS evaluates the defaults of `when`-hidden items and the `kots-helm.yaml` mapping only checks non-empty, so every non-GCS dev-licensed install got a GCS project id injected into `googleCloudStorageProjectId` — a field the operator can't even see, let alone override. Nothing reads it cross-provider (only the platform GCS storage client), so gating is safe.
2. `chart/templates/{import-api,import-worker}/configmap.yaml` — `WORKSPACE_THUMBNAILS_PROJECTID` / `WORKSPACE_THUMBNAILS_KEYFILENAME` now render only when `storageProvider` is `gcp`. These two templates gated the vars on value-emptiness / SA-key presence; `workspace-api/configmap.yaml` and `_commonChecks.tpl` already wrap the same block in the provider check — this aligns the two stragglers with the established pattern.

**Reverted from an earlier revision:** this PR originally also gated the `storageBucketsExportGcp` default the same way. That was wrong and is reverted here: `EXPORTS_GCS_BUCKET_NAME` is not GCS-provider-only config — maps-api / sql-worker use it as the BigQuery export staging bucket on **every** platform provider (selected per warehouse, coexisting with the AWS exports bucket by design), and the dev default points at a bucket that exists in dev tenant projects. Gating it would have emptied a live BQ-exports bucket on non-GCS dev installs. The Azure misleading-error symptom is instead being fixed app-side (cloud-native#28017 adds `EXPORTS_AZURE_CONTAINER_NAME` and rejects BQ exports that cannot stage through GCS); chart wiring for that var and decoupling the KOTS exports items from `storageBucketsBehaviour` are tracked as follow-ups.

**Benefits**

- S3 and Azure installs: configmaps no longer carry a GCS project id or GCS key paths that the running provider never reads — removing authoritative-looking dead config that derails storage debugging.

**Possible drawbacks**

None expected. GCS dev installs (`custom_gcs`) keep the default unchanged; non-dev licenses never received it. The thumbnails vars removed from s3/azure renders are unread on those providers (the storage clients select credentials per provider), so rendered output only loses dead keys.

**Follow-ups (not addressed here)**

- Wire `EXPORTS_AZURE_CONTAINER_NAME` (new in cloud-native#28017) in the chart + KOTS so Azure platform exports stop falling back to the GCS-named var.
- Decouple the KOTS exports items from `storageBucketsBehaviour`: the app supports a GCS exports bucket (BigQuery staging) and an AWS exports bucket (Snowflake/Redshift UNLOAD) coexisting on any platform storage, but the KOTS UI only ever shows the one matching the platform storage provider.
- `import-api` / `import-worker` resolve the exports container through their platform-storage client, while `maps-api` / `sql-worker` receive dedicated `EXPORTS_S3_BUCKET_*` credentials — on an S3-compatible deployment where the exports bucket does not live on the platform storage endpoint, import-api exports address a bucket their client cannot reach.

**Applicable issues**

- [sc-569088]

**Additional information**

Validated with `scripts/test-kots-config.sh all` (static checks only — YAML syntax and ConfigOption references; `default:` expressions are not evaluated outside a KOTS install; the `accessToCartoModeK8sNotFullySupported` warning is pre-existing on main) and `helm template` on both paths (plain and `--set replicated.enabled=true`). Render checks: with `storageProvider=s3` and a non-empty `googleCloudStorageProjectId`, `WORKSPACE_THUMBNAILS_PROJECTID` / `WORKSPACE_THUMBNAILS_KEYFILENAME` no longer render in import-api or import-worker; with `storageProvider=gcp` both components render them as before. The project-id default change was additionally verified on a live dedicated-k8s install from this PR's dev channel (empty on a dev s3 install, vs the leaked value on the released chart).


mateo-di:
### Verified live on dedicated-k8s

Installed a dev-licensed `custom_aws_s3` dedicated-k8s env from this PR's dev channel and compared it against an identical env on the released chart (same chart version) — this is the one check `helm template` and `scripts/test-kots-config.sh` can't cover, since KOTS `default:` expressions only evaluate at install time.

| `import-api` configmap var | released chart (no guard) | this PR |
|---|---|---|
| `EXPORTS_GCS_BUCKET_NAME` | `<gcp-project-id>-export-storage` | `""` |
| `WORKSPACE_THUMBNAILS_PROJECTID` | `<gcp-project-id>` | not rendered |
| `WORKSPACE_THUMBNAILS_KEYFILENAME` | GCS SA key path | not rendered |
| GCS-only vars in configmap | 3 | **0** |
| imports / thumbnails buckets | intact | intact |

Same result in `import-worker`. Because the released chart at this same version still carries the `dev`-only default, an empty `EXPORTS_GCS_BUCKET_NAME` on a non-GCS dev install can only come from this PR's added `custom_gcs` guard — so the behaviour change is attributable to this diff alone.

manifests/kots-config.yaml
```diff
@@ -513,7 +513,7 @@ spec:
           help_text: The GCP Project Id of these buckets.
           type: text
           required: true
-          default: repl{{ if ConfigOptionEquals "replicatedLicenseType" "dev"}}repl{{ ConfigOption "derivedSelfHostedGCPProjectId"}}repl{{ end }}
+          default: repl{{ if and (ConfigOptionEquals "replicatedLicenseType" "dev") (ConfigOptionEquals "storageBucketsBehaviour" "custom_gcs")}}repl{{ ConfigOption "derivedSelfHostedGCPProjectId"}}repl{{ end }}
```
moicalcob, line 516: Hey @mateo-di but this is not failing as we discussed no? 🤔 I mean, what we're setting is the default value, but when the user provides a different one it's working?? I don't get the issue...

moicalcob (changes requested):

manifests/kots-config.yaml
mateo-di: Hey @moicalcob, analyzing again the source of the issue with the recent Sajid's PRs fix in cloud-native and our chart setup for the exportsBucket wiring wasn't the responsable of this issue. So the wiring is ok!! The thing was how imports-exports code were using the Env Vars. The problem was how the imports/exports code was using the env vars: import-api reused EXPORTS_GCS_BUCKET_NAME as the Azure container name that was fixed in app-side in cloud-native#28017 with a dedicated EXPORTS_AZURE_CONTAINER_NAME.

in the dedicated-k8s envs the GCS exports bucket (dev-license default) and the AWS exports bucket (set explicitly by the dedicateds pipeline) coexist.  So the PR now only gates the unused storageBucketsGcpProjectId to custom_gcs (the default, plus the import-api/import-worker template renders of the thumbnails GCS vars). From what i saw nothing reads a GCS project id on other providers the BQ staging path only needs bucket + credentials.

mateo-di:
### Helm-path validation on a live install (new + upgrade)

Merged `main` (11 commits behind) and re-validated on a real standalone install — a GKE cluster with Gatekeeper enforcing, using a dev test-customer package whose storage provider is GCS, external Cloud SQL, 14/14 pods healthy.

**1. The leak, reproduced on the released chart.** Same customer package, same values, only the chart version differs — `storageProvider=s3`:

| component | released `1.295.1` | this PR |
|---|---|---|
| `import-api` | `WORKSPACE_THUMBNAILS_PROJECTID`, `WORKSPACE_THUMBNAILS_KEYFILENAME` | *(neither rendered)* |
| `import-worker` | `WORKSPACE_THUMBNAILS_PROJECTID`, `WORKSPACE_THUMBNAILS_KEYFILENAME` | *(neither rendered)* |
| `workspace-api` | *(neither rendered)* | *(neither rendered)* |

`workspace-api` being already clean on the released chart is the point — it has carried the `storageProvider == gcp` guard all along. This PR brings the two stragglers in line with it, it does not invent a new pattern.

**2. GCS installs are untouched.** `storageProvider=gcp`, this PR — all three components still render both vars:

```
import-api:      WORKSPACE_THUMBNAILS_PROJECTID, WORKSPACE_THUMBNAILS_KEYFILENAME
import-worker:   WORKSPACE_THUMBNAILS_PROJECTID, WORKSPACE_THUMBNAILS_KEYFILENAME
workspace-api:   WORKSPACE_THUMBNAILS_PROJECTID, WORKSPACE_THUMBNAILS_KEYFILENAME
```

**3. The reverted part stays reverted.** `EXPORTS_GCS_BUCKET_NAME` is warehouse-scoped, not provider-scoped, so it must keep rendering everywhere. Confirmed on all three providers with this PR:

| provider | `EXPORTS_GCS_BUCKET_NAME` |
|---|---|
| `gcp` | `<tenant>-export-storage` |
| `s3` | `<tenant>-export-storage` |
| `azure-blob` | `<tenant>-export-storage` |

**4. Live upgrade, not just `helm template`.** Upgraded the running install from released `1.295.1` to this PR's chart:

- upgrade to the PR chart on `gcp` → revision 3, both vars still present in the live ConfigMaps, 14/14 pods Running
- upgrade with `storageProvider=s3` → revision 4, both vars **gone** from the live `carto-import-api` / `carto-import-worker` ConfigMaps, while `WORKSPACE_THUMBNAILS_PROVIDER=s3` and `WORKSPACE_THUMBNAILS_REGION` render correctly
- back to `gcp` → revision 5, vars restored

So the behaviour holds on an upgrade of an existing install, not only on a fresh render.

---

@moicalcob — your change-request was about whether the wiring was really at fault. It wasn't, and that is already reverted in this PR: the `storageBucketsExportGcp` gating is gone, `EXPORTS_GCS_BUCKET_NAME` renders on every provider (table 3), and the Azure symptom is fixed app-side in cloud-native#28017. What's left here is only the two GCS-only leaks above plus the `storageBucketsGcpProjectId` dev-license default. Re-requesting review when you have a moment.

KOTS-path validation (the `default:` expression, which only evaluates at install time) is tracked separately in this PR's thread — the fresh dev-licensed install was verified earlier; the upgrade path is being covered now.

mateo-di:
### Leak reproduced on a live KOTS install (dev license, `custom_aws_s3`)

Brought up a fresh dedicated-k8s environment on the released `Dedicateds` channel (`carto-1.295.1` / `2026.9.4`) — a **dev-licensed KOTS install whose storage behaviour is `custom_aws_s3`**, i.e. exactly the configuration this PR is about. Reading the deployed ConfigMaps:

| component | released chart, `WORKSPACE_THUMBNAILS_PROVIDER=s3` |
|---|---|
| `carto-import-api` | `WORKSPACE_THUMBNAILS_PROJECTID=<gcp-project-id>`, `WORKSPACE_THUMBNAILS_KEYFILENAME=/usr/src/certs/gcp-buckets-service-account/key.json` |
| `carto-import-worker` | `WORKSPACE_THUMBNAILS_PROJECTID=<gcp-project-id>`, `WORKSPACE_THUMBNAILS_KEYFILENAME=/usr/src/certs/gcp-buckets-service-account/key.json` |
| `carto-workspace-api` | *(neither — only `WORKSPACE_THUMBNAILS_PROVIDER`)* |

That is the whole thesis of this PR reproduced on a real install: a GCS project id and a GCS SA key path sitting in the ConfigMaps of an **S3** deployment, in the two components that lacked the provider guard, while `workspace-api` — which has carried the guard all along — is clean. The fix aligns the two stragglers with the component that already does it right.

### The reverted half is also confirmed correct on this install

Both exports buckets are populated simultaneously on the same s3 install:

```
EXPORTS_GCS_BUCKET_NAME = <tenant>-export-storage            # BigQuery staging
EXPORTS_S3_BUCKET_NAME  = <dedicated-env>-export-storage     # Snowflake / Redshift staging
```

This is the concrete evidence for why gating `storageBucketsExportGcp` was wrong and had to be reverted: on a non-GCS install the GCS exports bucket is live and in use for BigQuery staging. Gating it would have emptied a working bucket. `EXPORTS_AZURE_CONTAINER_NAME` is absent here, which is what #934 adds.

@moicalcob this should answer the original question directly — the wiring was never at fault (both exports buckets coexist, as above), and what remains in this PR is only the two GCS-only vars leaking onto non-GCS providers plus the `storageBucketsGcpProjectId` dev-license default.

The upgrade-path run on this same environment (released -> this PR's channel) is next.

mateo-di:
### KOTS upgrade path verified — before/after on the same live install

Same dedicated-k8s environment as my previous comment (dev license, `custom_aws_s3`). Moved its customer from the released `Dedicateds` channel to this PR's channel and upgraded in place (`kots upstream upgrade`, sequence 2 -> 7). No reinstall, so this is genuinely the upgrade path an existing customer takes.

`WORKSPACE_THUMBNAILS_PROVIDER=s3` throughout:

| component | released chart (seq 2) | this PR (seq 7) |
|---|---|---|
| `carto-import-api` | `WORKSPACE_THUMBNAILS_PROJECTID`, `WORKSPACE_THUMBNAILS_KEYFILENAME` | *(gone — only `WORKSPACE_THUMBNAILS_PROVIDER`)* |
| `carto-import-worker` | `WORKSPACE_THUMBNAILS_PROJECTID`, `WORKSPACE_THUMBNAILS_KEYFILENAME` | *(gone — only `WORKSPACE_THUMBNAILS_PROVIDER`)* |
| `carto-workspace-api` | *(already clean)* | *(still clean)* |

And the reverted half holds on the upgrade too — both exports buckets remain populated on this non-GCS install:

```
EXPORTS_GCS_BUCKET_NAME = <tenant>-export-storage           # BigQuery staging
EXPORTS_S3_BUCKET_NAME  = <dedicated-env>-export-storage    # Snowflake / Redshift staging
```

### Coverage now complete

| quadrant | result |
|---|---|
| Helm, new install | GCS-only vars absent on `s3`/`azure-blob`, present on `gcp` |
| Helm, upgrade | released -> PR on a live install, 14/14 pods healthy, vars behave per provider |
| KOTS, fresh install (dev license, `custom_aws_s3`) | leak reproduced on the released chart |
| KOTS, upgrade from released | **table above** |

Merged `main` (11 commits), CI green.

@moicalcob this is ready for another look whenever you have time. Short version of the original question: the exports wiring was never the culprit — both exports buckets coexist by design, and the evidence above shows them both live on a non-GCS install, which is exactly why the `storageBucketsExportGcp` gating was reverted out of this PR. What remains is only the two GCS-only vars leaking into non-GCS ConfigMaps, plus the `storageBucketsGcpProjectId` dev-license default.

mateo-di:
### Clarification on the KOTS evidence above

Correcting my own wording, since an earlier comment here said "KOTS upgrade path verified" without saying how.

**What was done:** a live dev-licensed dedicated-k8s environment (`custom_aws_s3`), moved onto this PR's Replicated channel via `replicated customer update` and upgraded in place with `kubectl kots upstream upgrade` (sequence 2 → 7). The before/after ConfigMap tables I posted come from that, and they stand — the leak is real on the released chart and gone on this PR, on a genuine KOTS install.

**What was not done:**

1. A **fresh `kots install` from this PR's channel**. The channel override is only consumed at the first install, so every run after the environment existed took the update path.
2. An update driven by **commit → CI build → ded deploy**. Mine was a manual `kots upstream upgrade`, not the pipeline's `deploy-step: update`.

For this PR I think that residual is acceptable, and here's the specific reason rather than a hand-wave: the change is a KOTS `default:` expression plus two template guards. The `default:` behaviour is the part that only evaluates at install time — and I did observe it on a live dev-licensed `custom_aws_s3` install, which is exactly the configuration it targets. A fresh install would re-confirm the same expression against the same license type. The chart-side guards are provider conditionals, already covered on both the Helm render matrix and a live Helm upgrade.

Full stage coverage for this PR:

| stage | status |
|---|---|
| Fresh `dedicated-k8s` install from this PR's channel | not tested (pool is 35/35 reserved) |
| Update over an existing KOTS install | done — manual channel switch + `kots upstream upgrade` |
| Preflights in a non-deployed namespace | n/a — no preflight spec change |
| Upgrade of an existing Helm deployment | done — revisions 3/4/5, gcp → s3 → gcp, 14/14 pods |

@moicalcob still ready for your look — the substance of the fix and the evidence for it are unchanged, this is just me being precise about test provenance.

PR #934, opened 2026-08-31
feat(selfhosted): add the Azure exports container to chart and KOTS config [sc-569419]

mateo-di:
**Description of the change**

Wires the new `EXPORTS_AZURE_CONTAINER_NAME` (introduced app-side in cloud-native#28017) end to end, so Azure Blob deployments get a dedicated exports-container setting instead of overloading the GCS-named exports field:

- `chart/values.yaml` — new `appConfigValues.azureExportContainer` (default `""`), README regenerated.
- `chart/templates/{import-api,import-worker}/configmap.yaml` — render `EXPORTS_AZURE_CONTAINER_NAME`, unconditionally like its `EXPORTS_GCS_BUCKET_NAME` / `EXPORTS_S3_BUCKET_NAME` siblings.
- `manifests/kots-config.yaml` — new `storageBucketsExportAzure` item ("Azure exports container name"), shown for `custom_azb`, optional, no default (Azure containers are not auto-provisioned).
- `manifests/kots-helm.yaml` — maps it to `appConfigValues.azureExportContainer` when non-empty.

**Benefits**

- Azure installs can name their exports container through a field that says what it is, instead of a GCS-named one.
- KOTS Azure customers gain an exports option at all — until now no exports item was shown for `custom_azb`.

**Possible drawbacks**

None expected. The app resolves the Azure exports container as `EXPORTS_AZURE_CONTAINER_NAME ?? EXPORTS_GCS_BUCKET_NAME`, and its settings reader maps a blank env var to null — verified in source — so existing Azure installs configured through the GCS-named field keep working unchanged, and rendering the new var empty on other providers is inert.

**Applicable issues**

- [sc-569419]

**Additional information**

Follow-up tracked separately ([sc-571305]): decoupling the KOTS exports items from `storageBucketsBehaviour` so the per-warehouse exports buckets can coexist on any platform storage — an RFC PR accompanies it.

Validated with `scripts/test-kots-config.sh all` (static) and `helm template` on both paths (plain and `--set replicated.enabled=true`): with `storageProvider=azure-blob` and the container set, `EXPORTS_AZURE_CONTAINER_NAME` renders with the value; on other providers it renders empty alongside its siblings. `chart/README.md` regenerated via readme-generator-for-helm.


mateo-di:
### KOTS validation

Upgraded a live dev-licensed dedicated-k8s install onto this branch's channel (`kots upstream upgrade`, sequence 9 -> 11). That environment's storage behaviour is `custom_aws_s3`, so it exercises the **non-Azure** path — the one where this change must be completely inert.

`carto-import-api` and `carto-import-worker` after the upgrade:

```
EXPORTS_AZURE_CONTAINER_NAME =                                  # new, rendered empty
EXPORTS_GCS_BUCKET_NAME      = <tenant>-export-storage          # unchanged
EXPORTS_S3_BUCKET_NAME       = <dedicated-env>-export-storage    # unchanged
```

30/30 pods healthy through the upgrade. The new var renders empty (its KOTS item is hidden under `when: custom_azb`, so nothing is mapped), and both sibling exports vars are untouched — which is the required outcome for every existing non-Azure install.

### Why an empty value is safe for existing Azure installs

This was the one part of this PR with real breakage potential, since the chart renders the var **empty rather than absent**. Verified against the merged app-side code rather than assuming:

`packages/backend-utils/src/settings.ts` — `getStringSetting` with `{ optional: true }`:

```
if (typeof settingValue === 'string' && settingValue.trim().length === 0) {
  if (options?.optional) {
    return null
  }
```

A blank env var returns `null`, not `''`. And `import-api/src/settings.ts` resolves:

```
return getStringSetting('EXPORTS_AZURE_CONTAINER_NAME', { optional: true }) ??
  getStringSetting('EXPORTS_GCS_BUCKET_NAME', { optional: true })
```

`null ?? x` falls through, so an Azure install that configured its container through the GCS-named field keeps working. cloud-native#28017 also carries a regression test for exactly this — *"falls back to EXPORTS_GCS_BUCKET_NAME on azure-blob so existing deployments keep working"*.

### Coverage

| quadrant | result |
|---|---|
| Helm render, all providers | takes the value on `gcp`/`s3`/`azure-blob` when set; empty (present, not absent) when unset — same shape as its `EXPORTS_S3_BUCKET_NAME` sibling |
| Helm, both install paths | renders clean plain and with `replicated.enabled=true` |
| KOTS config validation | `scripts/test-kots-config.sh all` passes; item gated `when: custom_azb`, optional, no default |
| KOTS, upgrade from released (non-Azure) | table above — inert, no disruption |
| App-side contract | blank -> `null` -> falls back to the GCS var, verified in merged source + upstream test |

**Not covered:** a live `custom_azb` KOTS install where the item is visible and populated. The dedicated pool is GCS/S3-backed, so there is no Azure-storage environment available. That path rests on the static `when:` validation plus the render tests above rather than a live install — flagging it explicitly rather than implying full coverage.

Merged `main` (11 commits), CI green.

mateo-di:
### Stage 4: upgrade of an existing Helm deployment

Previously I had only shown `helm template` renders for this PR. Closing that gap — a real install upgraded in place, not a fresh render.

Baseline: released `carto/carto 1.295.1` installed with a dev test-customer package (GCS provider), external Cloud SQL, 14/14 pods healthy. Then `helm upgrade` onto this PR's chart.

| revision | chart | `EXPORTS_AZURE_CONTAINER_NAME` | `EXPORTS_GCS_BUCKET_NAME` | `EXPORTS_S3_BUCKET_NAME` | pods |
|---|---|---|---|---|---|
| 1 | released `1.295.1` | **absent** | `<tenant>-export-storage` | `""` | 14/14 |
| 2 | this PR | `""` (present, empty) | `<tenant>-export-storage` | `""` | 14/14 |
| 3 | this PR + `azureExportContainer=live-upgrade-container` | `live-upgrade-container` | `<tenant>-export-storage` | `""` | 14/14 |

Both `carto-import-api` and `carto-import-worker` behaved identically at every revision.

That covers the three things that mattered for an upgrade:

1. **The var appears where it didn't exist before** (rev 1 → 2) without disturbing its siblings.
2. **Present-but-empty, not absent** — the property the app-side `EXPORTS_AZURE_CONTAINER_NAME ?? EXPORTS_GCS_BUCKET_NAME` fallback depends on. Verified earlier against merged `getStringSetting`, which returns `null` (not `""`) for a blank value.
3. **The value actually wires through on an upgrade** (rev 3), not just on a fresh render.

No pod restarts beyond the rolling update, and no revision left the install unhealthy.

### Test-stage coverage for this PR

| stage | status |
|---|---|
| Fresh `dedicated-k8s` install from this PR's channel | **not tested** — see note below |
| Update over an existing KOTS install | done, but by manual channel switch + `kots upstream upgrade`, **not** via commit → CI → ded deploy |
| Preflights in a non-deployed namespace | n/a — this PR changes no preflight spec |
| Upgrade of an existing Helm deployment | **done, table above** |

Being explicit because my earlier comment said "KOTS validation" without qualifying it: the KOTS check I ran was a manual `replicated customer update` + `kots upstream upgrade` against a live dev-licensed env. It exercises the KOTS render and the upgrade path, but it is **not** the CI pipeline's `deploy-step: update`, and **no fresh install from this PR's channel has been done**. For this PR I think that is an acceptable residual — the KOTS item is `when: custom_azb` and the mapping is statically validated, and a fresh install would mainly re-confirm the same render. Flagging it rather than leaving it implied.

PR #938, opened 2026-09-03
fix(chart): route ai-api backend calls through the router for MCP usage metering [sc-571540]

moicalcob:
**Description of the change**

Self-Hosted half of the MCP usage-metering fix ([sc-571540](https://app.shortcut.com/cartoteam/story/571540), design in the Spec-Dev via CartoDB/cloud-native#28251). ai-api reaches its backends through in-cluster Service URLs, so the `CM` usage headers the backends emit never pass `router-http`'s access log and MCP-driven work (imports, queries, maps, LDS) generates no consumption.

The four billable-SKU `AI_*_API_URL` now point at the router, unconditionally — there is no metered path that doesn't go through it:

- `AI_SQL_API_URL` / `AI_MAPS_API_URL` / `AI_LDS_API_URL` / `AI_IMPORT_API_URL` → `<router-internal>/api` (the router strips `/api`; ai-api already builds `/v3/...` paths)
- `AI_WORKSPACE_API_URL` stays on the direct workspace-api Service — same rule as SaaS: only the billable SKUs route through the metering edge (workspace is rate-zero, so routing it meters nothing)

The target is a new **`<router>-internal` ClusterIP Service** exposing a new plain-HTTP nginx listener (`ROUTER_INTERNAL_HTTP_PORT`, added to the router image in CartoDB/cloud-native#28251; container port `router.containerPorts.internalHttp`, default 8085). Rationale: with `tlsCerts.httpsEnabled` (the default) the router's public HTTP port only 301-redirects, and its HTTPS cert names the external domain — while calling `https://<selfHostedDomain>` from inside the cluster would depend on hairpin DNS and on cert trust that `tlsCerts.autoGenerate` (the default) makes impossible to pin (its CA is re-minted on every render). The internal listener works identically in every TLS configuration and is deliberately **not** added to the public Service, so it can never bypass edge TLS from outside.

`AI_ACCOUNTS_API_URL` / `AI_DO_CATALOG_API_URL` are untouched: they point at CARTO's SaaS-hosted domains (`cartoAccApiDomain` / `cartoDoApiDomain`), so there is no local metering edge for them — this resolves the Spec-Dev's Open Question 15.

**Benefits**

MCP-driven API usage produces `request_metrics` rows via router-metrics, exactly like Workspace/CLI traffic — closing the under-billing and cap-bypass gap on Self-Hosted from the release that ships this chart. No TLS, DNS, or cert-trust prerequisites for any install.

**Possible drawbacks**

- Requires a router image that includes `ROUTER_INTERNAL_HTTP_PORT` support (CartoDB/cloud-native#28251) — this PR must ship in a release whose router image carries that change. On an older image the env var is ignored and ai-api's backend calls fail against the internal service; release pairing covers this.
- One extra in-cluster proxy hop per ai-api backend call.

**Applicable issues**

- Story: [sc-571540](https://app.shortcut.com/cartoteam/story/571540)
- Depends on: CartoDB/cloud-native#28251 (router internal listener + Spec-Dev Decision 17 / Phase 15 + the SaaS terraform half)

**Additional information**

Validated with `helm template` (ai-api configmap, router configmap/deployment, internal Service all render as intended) and `helm lint`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


moicalcob:
## Validated on ded-k8s-04 (dedicated selfhosted, CartoDB/cloud-native#28251's env) ✅

Applied this PR's chart changes to the env (router image from the cloud-native PR branch, which carries the `ROUTER_INTERNAL_HTTP_PORT` listener) and drove MCP tool calls with an API Access Token.

**Plumbing:**
- `carto-router-internal` ClusterIP Service up, targeting the named `internal-http` port (8085); the port is absent from the public Service
- ai-api env: `AI_SQL/MAPS/LDS/IMPORT_API_URL → http://carto-router-internal.dedicated-28251.svc.cluster.local/api`, `AI_WORKSPACE_API_URL → …/workspace-api`; accounts/do-catalog unchanged
- In-cluster probes through the listener: `/api/v3/sql/...` → 401, `/workspace-api/connections` → 401 (routed to the right backends, auth enforced)

**Metering — the router access log (source of `router-metrics` → `aggregated_selfhosted_metrics`) now observes MCP-driven usage:**

```
15:53:29 200 sql_api_sync_req_carto-mcp-server-unknown        /api/v3/sql/carto_dw/query       ac_6i1ld5mj
15:53:30 502 lds_api_req_carto-mcp-server-unknown             /api/v3/lds/geocoding/geocode    ac_6i1ld5mj
15:53:31 200 maps_api_rev202404_req_carto-mcp-server-unknown  /api/v3/maps/carto_dw/query      ac_6i1ld5mj
```

(host on every row: `carto-router-internal.dedicated-28251.svc.cluster.local` — the geocode 502 is the env's LDS provider being unconfigured; the metering row is what matters, and errors are observed like any client's.)

Before this change these calls hit the backend Services directly and produced no log line at all. Import wasn't exercised here (the test token lacks the Imports API) — that surface was validated end-to-end on the SaaS ded env (see CartoDB/cloud-native#28251). The `router-metrics` sidecar is healthy and flushes aggregates to PubSub on its 5-minute window.

Notes from the test:
- MCP tools also worked through the router with no functional regression (query returned rows, describe returned schema)
- The caller-adaptive surface behaves correctly through the metered path: a token without the LDS/Imports APIs doesn't get those tools

chart/templates/router/service-internal.yaml
```diff
@@ -0,0 +1,27 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: router
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  type: ClusterIP
```
mateo-di: Separate Service is the right call, but not quite for the reason in the description. `router.service.type` defaults to ClusterIP and the ingress binds the backend by port *name* (`servicePort: "http"`), so the main Service isn't publicly reachable either. What this actually protects is an install that overrides the type to LoadBalancer/NodePort, where every port on the Service gets exposed. Worth stating that case explicitly, otherwise someone later "simplifies" this into `router.service.extraPorts` and quietly loses it.

chart/templates/router/service-internal.yaml
```diff
@@ -0,0 +1,27 @@
...
+  ports:
+    - port: 80
```
mateo-di, line 17: Any reason this is hardcoded here? Every other port in this chart is a value (`router.service.ports.http`). Same for the Service itself having no enabled flag.

chart/templates/router/configmap.yaml
```diff
@@ -51,6 +51,7 @@ data:
   ROUTER_WORKSPACE_API_INTERNAL_URL: {{ include "carto.workspaceApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
   ROUTER_WORKSPACE_WWW_INTERNAL_URL: {{ include "carto.workspaceWww.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
   ROUTER_ENABLE_HTTPS: {{ .Values.tlsCerts.httpsEnabled | quote }}
+  ROUTER_INTERNAL_HTTP_PORT: {{ .Values.router.containerPorts.internalHttp | quote }}
```
mateo-di: No guard, so the internal listener renders in every install — including `tlsCerts.httpsEnabled: false`, where the main server block already listens on plain HTTP 8080 and this is just a duplicate listener.

I'd keep it uniform (one path in every TLS config is the point), but it does mean cloud-native#28251's "unset by default so existing deployments render byte-identical config" only holds for the image, not for any chart install.

chart/templates/ai-api/configmap.yaml
```diff
@@ -17,13 +17,12 @@ data:
...
-  AI_SQL_API_URL: "http://{{ include "carto.mapsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_MAPS_API_URL: "http://{{ include "carto.mapsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_LDS_API_URL: "http://{{ include "carto.ldsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_WORKSPACE_API_URL: "http://{{ include "carto.workspaceApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_IMPORT_API_URL: "http://{{ include "carto.importApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
+  {{- $routerBase := printf "http://%s-internal.%s.svc.%s" (include "carto.router.fullname" .) .Release.Namespace .Values.clusterDomain }}
+  AI_SQL_API_URL: "{{ $routerBase }}/api"
+  AI_MAPS_API_URL: "{{ $routerBase }}/api"
+  AI_LDS_API_URL: "{{ $routerBase }}/api"
+  AI_WORKSPACE_API_URL: "{{ $routerBase }}/workspace-api"
```
mateo-di, line 24: Same as in the other review to clarify! cloud-native PR keeps `AI_WORKSPACE_API_URL` pointing straight at workspace-api on SaaS (rate zero SKU, and Decision 17 says anything we dont meter stays direct). Here it goes through the router instead. Whats the reason? Intentional divergence between the two models, or should this match the SaaS side?

chart/templates/ai-api/configmap.yaml
```diff
@@ -17,13 +17,12 @@ data:
...
   # aiProxy acts as a unified gateway that can route to various AI providers
   AI_OPENAI_ENDPOINT: "http://{{ include "carto.aiProxy.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}/v1"
-  # AI_SQL_API_URL and AI_MAPS_API_URL both point to maps-api: it serves the SQL
-  # and the Maps API surfaces in self-hosted, unlike SaaS where they are split
-  AI_SQL_API_URL: "http://{{ include "carto.mapsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_MAPS_API_URL: "http://{{ include "carto.mapsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_LDS_API_URL: "http://{{ include "carto.ldsApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_WORKSPACE_API_URL: "http://{{ include "carto.workspaceApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-  AI_IMPORT_API_URL: "http://{{ include "carto.importApi.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
+  {{- $routerBase := printf "http://%s-internal.%s.svc.%s" (include "carto.router.fullname" .) .Release.Namespace .Values.clusterDomain }}
```
mateo-di: Old router image + this chart = nginx never binds 8085 and every ai-api backend call gets connection refused. That's all AI features down, not just metering, and `router.image.tag` is overridable (support pins images). Release pairing covers the happy path only — is there anything that fails loudly, or one value to flip these five back to the direct Services? That lever isn't a metering opt-out, it's a different axis from Decision 17's "no toggle".

Also: the old comment about sql/maps both being maps-api in self-hosted goes away here, and that asymmetry now only exists in the router config.

mateo-di (dismissed):
LGTM! With some clatifications

chart/templates/router/service-internal.yaml
moicalcob: Good catch on the convention — the service port is now `router.service.ports.internalHttp` (default 80), and the targetPort was already the values-driven named container port (`router.containerPorts.internalHttp`). On the enabled flag: intentional — Decision 17 in the Spec-Dev makes the metering re-point unconditional (there is no valid configuration where MCP usage should bypass metering), so the Service ships always, same as the routing itself. An earlier iteration had a toggle and it was deliberately removed.

chart/templates/ai-api/configmap.yaml
moicalcob: Update — we've aligned it with SaaS instead: `AI_WORKSPACE_API_URL` now stays on the direct workspace-api Service here too, so both models follow one rule: only the billable SKUs (sql, maps, lds, import) route through the metering edge; workspace bills nothing and moves no cap, so routing it bought no metering. The trade-off (MCP workspace ops absent from activity data on both models, accepted) is recorded in the Spec-Dev's Decision 17. Thanks for pushing on this — the previous asymmetry wasn't worth its explanation.

mateo-di (dismissed):
Nice one! 💯 

PD: Check the readme please ⚠️

mateo-di (approved):

